PR #13, opened 2024-06-06
Refactor per region lookup

phw:
This refactors the multi-region lookups in Tidal and iTunes to use a common abstraction.

It implements abstract base classes `MetadataApiProvider` and `ReleaseApiLookup` as specializations of the existing abstract classes that hold functionality common to typical API based providers. 

`MetadataApiProvider`  for now only enforces the implementation of a `query` function as we already had it on the Deezer, iTunes and Tidal providers. `ReleaseApiLookup` provides the helper method `queryAllRegions` which generalizes performing release queries for each region until a result is found.

`queryAllRegions` expects a validator callback, which must verify the request response. If the result is usable the validator must return true. If it returns false the next region is tried. There is a optional error validator which can be used to validate exceptions being thrown during query.

Overall this greatly simplifies the request handling in both the iTunes and Deezer provider.

The new base classes will also be suited for adding abstractions for e.g. access token handling a currently present in the Tidal provider.

### DIFF
--- a/providers/base.ts
+++ b/providers/base.ts
@@ -401,6 +401,4 @@ export type CacheEntry<Content> = {
 	timestamp: number;
 	/** Indicates that the cache entry has been created by the returning method. */
 	isFresh: boolean;
-	/** Successfully queried region of the API. */
-	region?: CountryCode;
 };
--- a/providers/base.ts
+++ b/providers/base.ts
@@ -1,5 +1,5 @@
 import { FeatureQuality, type FeatureQualityMap, type ProviderFeature } from './features.ts';
-import { ProviderError } from '@/utils/errors.ts';
+import { ProviderError, ResponseError } from '@/utils/errors.ts';
 import { getLogger } from 'std/log/get_logger.ts';
 import { rateLimit } from 'utils/async/rateLimit.js';
 import { simplifyName } from 'utils/string/simplify.js';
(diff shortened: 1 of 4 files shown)


providers/base.ts
```diff
@@ -351,6 +351,48 @@ export abstract class ReleaseLookup<Provider extends MetadataProvider, RawReleas
 	}
 }
 
+/** Extends MetadataProvider with functions common to lookups accessing web APIs. */
+export abstract class MetadataApiProvider extends MetadataProvider {
+	/** Must be implemented to perform a request against the specified URL. */
+	abstract query<Data>(apiUrl: URL, maxTimestamp?: number): Promise<CacheEntry<Data>>;
+}
```
kellnerd, line 358: Good idea to make this an explicit requirement, almost all providers had this anyway.

providers/base.ts
```diff
@@ -351,6 +351,48 @@ export abstract class ReleaseLookup<Provider extends MetadataProvider, RawReleas
...
+/** Extends ReleaseLookup with functions common to lookups accessing web APIs. */
+export abstract class ReleaseApiLookup<Provider extends MetadataApiProvider, RawRelease>
+	extends ReleaseLookup<Provider, RawRelease> {
+	/** Constructs an API URL for a release using the specified lookup options. */
+	abstract constructReleaseApiUrl(): URL;
+
+	/** Performs the query for the URL returned by constructReleaseApiUrl for all configured regions. */
+	protected async queryAllRegions<Data>(
+		dataValidator: (data: Data) => boolean,
+		errorValidator: (error: unknown) => boolean = (_) => false,
```
kellnerd, line 369: Without reading the implementation, I have an intuition of the meaning of `true` and `false` for the data validator, but not for the error validator.
Making the names more verbose might help:
```suggestion
		dataValidator: (data: Data) => boolean,
		ignoreError: (error: unknown) => boolean = (_) => false,
```
Or with inverse meaning:
```suggestion
		isValidData: (data: Data) => boolean,
		isCriticalError: (error: unknown) => boolean = (_) => true,
```

providers/base.ts
```diff
@@ -351,6 +351,48 @@ export abstract class ReleaseLookup<Provider extends MetadataProvider, RawReleas
 	}
 }
 
+/** Extends MetadataProvider with functions common to lookups accessing web APIs. */
```
kellnerd, line 354:
```suggestion
/** Extends `MetadataProvider` with functions common to lookups accessing web APIs. */
```

providers/base.ts
```diff
@@ -351,6 +351,48 @@ export abstract class ReleaseLookup<Provider extends MetadataProvider, RawReleas
...
+	abstract query<Data>(apiUrl: URL, maxTimestamp?: number): Promise<CacheEntry<Data>>;
+}
+
+/** Extends ReleaseLookup with functions common to lookups accessing web APIs. */
+export abstract class ReleaseApiLookup<Provider extends MetadataApiProvider, RawRelease>
+	extends ReleaseLookup<Provider, RawRelease> {
+	/** Constructs an API URL for a release using the specified lookup options. */
+	abstract constructReleaseApiUrl(): URL;
+
+	/** Performs the query for the URL returned by constructReleaseApiUrl for all configured regions. */
```
kellnerd, line 366:
```suggestion
	/** Performs the query for the URL returned by {@linkcode constructReleaseApiUrl} for all configured regions until valid data is returned. */
```
Untested, but this should make IDE tooltips link to the method (in VS Code).

providers/base.ts
```diff
@@ -351,6 +351,48 @@ export abstract class ReleaseLookup<Provider extends MetadataProvider, RawReleas
 	}
 }
 
+/** Extends MetadataProvider with functions common to lookups accessing web APIs. */
+export abstract class MetadataApiProvider extends MetadataProvider {
+	/** Must be implemented to perform a request against the specified URL. */
+	abstract query<Data>(apiUrl: URL, maxTimestamp?: number): Promise<CacheEntry<Data>>;
+}
+
+/** Extends ReleaseLookup with functions common to lookups accessing web APIs. */
```
kellnerd, line 360:
```suggestion
/** Extends `ReleaseLookup` with functions common to lookups accessing web APIs. */
```

kellnerd (commented):
Nice clean-up, so my initial fear that contributors would only be interested in writing providers and leave the refactoring to me was unjustified 😀

Looking good, but I haven't tested yet.

providers/base.ts
phw: Indeed it does!

kellnerd (approved):
I have carefully compared the old implementation with the new one for the case of the iTunes provider and found no relevant difference.

The new solution is much cleaner and easy to reuse while being less code, so definitely a good move.
We can even remove the `CacheEntry.region` attribute again which was kind of a hack to let the iTunes provider class return the region to the release lookup class, but I can't create a suggestion for this since you haven't touched that line.

phw:
> We can even remove the `CacheEntry.region` attribute again which was kind of a hack to let the iTunes provider class return the region to the release lookup class, but I can't create a suggestion for this since you haven't touched that line.

I removed it. The part of the code where the iTunes provider set this attribute on the cache entry looked indeed out of place. But now I understand better why it was there.

PR #18, opened 2024-06-07
fix(web): do not fail parsing country codes with surrounding spaces

phw:
If as a user you entered "GB, US" instead of "GB,US" this caused an error "' US' is not a valid ISO 3166-1 alpha-2 country code with two letters".

Be more tolerant with the input and trim the values after splitting by comma.


kellnerd (approved):
Always good to have input trimming, thanks. I don't think I have seen this specific error in the logs so far but it happened for GTIN, so that case had already been fixed.

PR #20, opened 2024-06-08
feat(Bandcamp): set creative commons license URL if available

### LINKED ISSUE #19: Support setting CC license for Bandcamp provider
Bandcamp contains many Creative Commons licensed releases, see for example https://aeonsable.bandcamp.com/album/aenigma-2023

The Bandcamp importer user script supports reading the license information and sets the license URL when seeding, see https://github.com/murdos/musicbrainz-userscripts/blob/master/bandcamp_importer.user.js#L188-L195

Similar could be done in the Bandcamp provider.

See https://community.metabrainz.org/t/harmony-music-metadata-aggregator-and-musicbrainz-importer/698641/12

phw:
This extracts Creative Common license URLs from Bandcamp pages and submits them as license URL link to MB.

This also generally completes the support for external links of type "license". Fixes #19

Some examples for testing:

- https://aeonsable.bandcamp.com/album/aenigma-2023
- https://aneagleinyourmind.bandcamp.com/album/intersection
- https://deusexlumina.bandcamp.com/track/my-church-is-black-me-and-that-man
- https://binaerpilot.bandcamp.com/album/songs-for-alan-turing

### DIFF
--- a/harmonizer/types.ts
+++ b/harmonizer/types.ts
@@ -108,7 +108,8 @@ export type LinkType =
 	| 'free streaming'
 	| 'mail order'
 	| 'paid download'
-	| 'paid streaming';
+	| 'paid streaming'
+	| 'license';
 
 /** MusicBrainz medium formats (incomplete). */
 export type MediumFormat =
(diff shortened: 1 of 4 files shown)


phw:
As a side note I think it might be helpful to actually handle the HTML response with a HTML parser and be able to use CSS path queries on it. Would lead to more robust code than the current regular expression based approach.

kellnerd (approved):

PR #27, opened 2024-06-10
fix(Tidal): fixed request error detection

phw:
This is a small fix for 6124885c41858e49b9e47f3acd5aa313856bacdb . The error handling detected all (non-empty) responses as error. Actually check for the `errors` attribute.


kellnerd (approved):
Oops, I caught and fixed this in the catch statement, then realized that the type assertion here was wrong but forgot to also update the condition 😬
Thanks for catching that!

Issue #6, opened 2024-05-16
Initial Roadmap


kellnerd:
This is just a loosely ordered list of things I already have on my radar, to be cleaned up later™️.

## Harmonized Data

- [x] Merge missing properties into the preferred provider's data
- [ ] Check for conflicting properties during merge 
  - [ ] Warning for duration (#136)
  - [x] Error for GTIN
  - [x] Error for incompatible medium track counts
  - [ ] Merge tracks if the total track count matches (1 medium vs N media)
  - [x] Skip missing tracklist! (by far the most common error in the test phase logs)
  - [ ] Merge empty medium into medium with tracks
- [ ] Release date quality ranking, plausibility checks for each provider (new attribute "date.warning"), merge strategy "prefer latest" (#101)
- [ ] Generally warn about pre-release data
- [ ] Guess featured artists from titles (#39)
- [x] Copyright notices
- [ ] Explicitness of tracks
- [ ] Optional title cleanup (numeric prefix, ETI style etc.)
  - [ ] Deezer allows crediting the same artist multiple times (#183)
- [ ] Customizable search & replace rules (#130)
- [ ] Audiobook / audio drama mode?
- [x] Preserve catalog numbers
- [ ] Improve language detection (skip too short inputs, try alternatives?)

## Providers

- [x] iTunes: Ensure that collection.trackCount equals the number of returned tracks
- [x] iTunes: Warn about responses which contain multiple release variants for an UPC
- [x] iTunes: Country list for optional region lookups
- [x] iTunes: Use region from URL
- [ ] iTunes: Try to use artist (or ISRC?) region for canonical (region-specific) URL
- [ ] iTunes: Try next region instead of throwing if JSON parsing fails (see screenshot)
- [x] Spotify: Pad GTIN with zeros if no results are found ([example](https://gist.github.com/atj/a244901c5d6b49b693b5302938555070))
- [x] Deezer: Truncate padded GTIN
- [x] Bandcamp: Add untitled hidden tracks, only their count is available as OG meta header
  - [x] Extract more than just trAlbum into snapshots -> wrapper object, avoid deserializing JSON
- [x] Bandcamp: Try band URL as label URL, extract label from packages
- [x] Bandcamp: Check whether band is part of the release artist before using it as label
- [x] iTunes: Warn about missing tracklist
- [x] Bandcamp: VA releases
- [x] Bandcamp: `/track` URLs (#7)
[...]

PR #29, opened 2024-06-11
fix(providers): warn if GTIN lookup found multiple results for Spotify / Tidal

### LINKED ISSUE #25: Barcode collision
Sometimes barcodes are not as unique as they should be...

635669065024 returns two different releases (different artists, but same label) with the iTunes, Spotify and Tidal providers. Deezer's API only returns one of them (YBC III), for the others it seems to be random which one is the first result that gets returned.

The iTunes provider at least warns about this, the other providers currently ignore this issue silently.

phw:
For Spotify and Tidal show a warning with a list of URLs if a GTIN lookup found more than one match.

Adapt the existing logic already implemented in iTunes and generalize it by adding a `ReleaseLookup.warnMultipleResults` helper.

Notes:
- Deezer seems to be technically limited to one result
- I don't know about Beatport. Maybe something similar should be done there

This should resolve #25

### DIFF
--- a/providers/Spotify/mod.ts
+++ b/providers/Spotify/mod.ts
@@ -173,8 +173,12 @@ export class SpotifyReleaseLookup extends ReleaseApiLookup<SpotifyProvider, Albu
 					this.constructReleaseApiUrl(),
 					this.options.snapshotMaxTimestamp,
 				);
-				if (cacheEntry.content?.albums?.items?.length) {
-					return cacheEntry.content.albums.items[0].id;
+				const releases = cacheEntry.content?.albums?.items;
+				if (releases?.length) {
+					if (releases.length > 1) {
+						this.warnMultipleResults(releases.slice(1).map((release) => release.external_urls?.spotify));
+					}
+					return releases[0].id;
 				}
 			}
 		}
--- a/providers/Tidal/mod.ts
+++ b/providers/Tidal/mod.ts
@@ -157,6 +157,9 @@ export class TidalReleaseLookup extends ReleaseApiLookup<TidalProvider, Album> {
 				return Boolean(data?.data?.length);
 			};
 			const result = await this.queryAllRegions<Result<Album>>(isValidData);
+			if (result.data.length > 1) {
+				this.warnMultipleResults(result.data.slice(1).map((release) => release.resource.tidalUrl));
+			}
 			return result.data[0].resource;
 		} else {
 			const isValidData = (data: Resource<Album>) => {
--- a/providers/base.ts
+++ b/providers/base.ts
@@ -21,6 +21,7 @@ import type { PartialDate } from '@/utils/date.ts';
 import type { CacheOptions, Snapshot, SnapStorage } from 'snap-storage';
 import type { MaybePromise } from 'utils/types.d.ts';
 import type { Logger } from 'std/log/logger.ts';
+import { pluralWithCount } from '@/utils/plural.ts';
 
 export type ProviderOptions = Partial<{
 	/** Duration of one rate-limiting interval for requests (in ms). */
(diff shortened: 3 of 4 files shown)


phw:
Some thoughts about this in general: In the long term it would be good if there could be some handling of multiple results. Something like this is already mentioned in the roadmap (#6) "Allow providers to return multiple releases, i.e. different variants (e.g. for Bandcamp)".

While testing I had one release that failed the lookup for iTunes due to mismatched track length. However, iTunes found multiple results and one that would have matched the tracklists from other providers was among them. I unfortunately could not find the example again in my notes, but if I do I'll post it.

providers/Spotify/mod.ts
```diff
@@ -173,8 +173,12 @@ export class SpotifyReleaseLookup extends ReleaseApiLookup<SpotifyProvider, Albu
 					this.constructReleaseApiUrl(),
 					this.options.snapshotMaxTimestamp,
 				);
-				if (cacheEntry.content?.albums?.items?.length) {
-					return cacheEntry.content.albums.items[0].id;
+				const releases = cacheEntry.content?.albums?.items;
+				if (releases?.length) {
+					if (releases.length > 1) {
+						this.warnMultipleResults(releases.toSpliced(0, 1).map((release) => release.external_urls?.spotify));
```
kellnerd, line 179:
```suggestion
						this.warnMultipleResults(releases.slice(1).map((release) => release.external_urls?.spotify));
```
I had to lookup what `toSpliced` is actually doing, it is a quite new method which does the same as a simple `slice` can do in this case.

providers/Tidal/mod.ts
```diff
@@ -157,6 +157,9 @@ export class TidalReleaseLookup extends ReleaseApiLookup<TidalProvider, Album> {
 				return Boolean(data?.data?.length);
 			};
 			const result = await this.queryAllRegions<Result<Album>>(isValidData);
+			if (result.data.length > 1) {
+				this.warnMultipleResults(result.data.toSpliced(0, 1).map((release) => release.resource.tidalUrl));
```
kellnerd, line 161:
```suggestion
				this.warnMultipleResults(result.data.slice(1).map((release) => release.resource.tidalUrl));
```

providers/base.ts
```diff
@@ -335,6 +336,20 @@ export abstract class ReleaseLookup<Provider extends MetadataProvider, RawReleas
 		};
 	}
 
+	/**
+	 * Shows a warning if the provider found more than the expected result for a lookup.
+	 * Expects a list of URLs pointing to the extra results from the provider.
+	 */
+	protected warnMultipleResults(urls: string[] | URL[]) {
+		const lines = urls.map((url) => `${url} ([lookup](/release?url=${encodeURIComponent(String(url))}))`);
```
kellnerd, line 344: I like the idea of having a link to initiate a new lookup, it is something which I had also considered doing. My only concern is that the provider implementation now does assumptions on how it is being used, that is how a host-relative link can initiate a lookup on the website.
That is something which I have tried to avoid so far. But I think we can temporarily add this until the providers have a dedicated way to indicate they have multiple results to select from.

Maybe we could use a document-relative link instead which only makes assumptions about the accepted query parameters?
```suggestion
		const lines = urls.map((url) => `${url} ([lookup](?url=${encodeURIComponent(String(url))}))`);
```

kellnerd (commented):
Thanks, that was fast!

kellnerd (approved):

kellnerd:
> While testing I had one release that failed the lookup for iTunes due to mismatched track length. However, iTunes found multiple results and one that would have matched the tracklists from other providers was among them. I unfortunately could not find the example again in my notes, but if I do I'll post it.

I have found the following example in my notes, taken from the server logs:
iTunes returns two totally unrelated releases by different artists: https://harmony.pulsewidth.org.uk/release?gtin=781676664221&itunes=&region=GB&ts=1717878974
Has a different number of tracks also, so it is not your example, but interesting nevertheless.

atj:
> * I don't know about Beatport. Maybe something similar should be done there

Just to clarify, the Beatport provider explicitly returns the first release from the search results with a matching barcode:

https://github.com/kellnerd/harmony/blob/0c9cd5f51a4fbf7cd60e08fdc94ec26f02206537/providers/Beatport/mod.ts#L106

I don't believe I've ever encountered a situation where a GTIN search has returned multiple releases, but the code will either return a single release or `undefined` regardless.

PR #43, opened 2024-06-21
Reword release action "Open in MusicBrainz"

MonkeyDo:
The smallest of typos, but I will surely not be the only one to notice.

### DIFF
--- a/server/routes/release/actions.tsx
+++ b/server/routes/release/actions.tsx
@@ -113,7 +113,7 @@ export default defineRoute(async (req, ctx) => {
 						<SpriteIcon name='brand-metabrainz' />
 						<p>
 							<a href={releaseUrl.href} target='_blank'>
-								Open on MusicBrainz
+								Open in MusicBrainz
 							</a>
 						</p>
 					</div>


mwiencek:
I'll be honest that I'd have also written this as "on" (like "watch on YouTube" or "search on Google") but I think you're right that "in" is more correct. 😁

kellnerd:
Indeed this is not really a typo but one case where I couldn't decide what is (more) correct.
I have been comparing this with related phrases I was more confident about ("Open (this release) in the app" versus "Open (this release) on the website/musicbrainz.org"), but in the end I flipped a coin.
After thinking more about it, I also err on the side of "in", but I would like to hear the opinion of our other native speakers @atj and @aerozol here just to be sure :)
Or maybe someone has a suggestion for a better wording to avoid having to make a call here?

Aerozol:
The English language strikes again...
I did a short browse of grammar websites etc, but I still don't really know. It seems that when referring to websites you should use "on". However this is referring to opening a specific item within a website, so "in" feels more natural to me as well.
In other words... who knows :P

kellnerd (approved):
Sorry that reviewing this minor change took so long, I had even managed to introduce a merge conflict for a one liner, oops 😅

Issue #23, opened 2024-06-09
Normalize and merge copyright lines


kellnerd:
Continuing the discussion from https://github.com/kellnerd/harmony/issues/22#issuecomment-2156460070

> We can factor out the copyright normalization logic and reuse it for other providers, e.g. as suggested for Tidal in https://community.metabrainz.org/t/harmony-music-metadata-aggregator-and-musicbrainz-importer/698641/15
> But I'd do this after merging this PR. It also needs some further research. I know Tidal includes the copyright text both with and without the © symbol. What I'm unsure is whether this strictly contains copyright © info, or whether it also can sometimes contain phonographic copyright ℗ info. Spotify has those separated, which makes it easier.

I fully agree, this is enough for its own PR and it needs more research.
Tidal also has a `copyright` property at the track level by the way, this should also be considered if it is different from the release level coypright. So far they were identical for the releases which I have checked, maybe a compilation has different values there.

For starters I have a [commit](https://github.com/kellnerd/harmony/commit/36b8d28676c1e6ccb46469cbf0a6273dfda7f3b3) in the `dev` branch which displays the alternative copyright values.
When we have more examples we can decide how the release merge algorithm should handle these, one possibility would be to keep all and deduplicate them.

phw:
For Tidal it is more complicated, the `copyright` field can contain both (P) or (C) entries, and it can be with or without a symbol (ASCII or proper character).

- This example has "(P) 2016 Century Media Records Ltd.": https://tidal.com/album/64944658
- Here it is "© 2012 S. Carter Enterprises, LLC.": https://tidal.com/browse/album/130201923
- And this is just "Nuclear Blast" without any symbol: https://tidal.com/browse/album/180584784

So I'm not sure whether we can or should add © or ℗ (or maybe "© + ℗") to the string if the symbol is missing. What we could do is converting the ASCII variants into symbols.

phw:
I wonder that maybe "copyright" should be a list of strings per provider instead of a single string. Currently it is I think only Spotify which offers explicit distinct values for (P) and (C). But having those as actual separate values will make de-duplication easier. Right now the provider just separates both with a new line.

SheamusPatt:
Why are you putting copyright / phonographic copyright in the annotation at all? There are release relationships where they can be put, which would link the release to the associated label or artist.
It would require matching the text after the statement to an artist or label (could be either).
That would apply to phonograpic copyright on tracks as well.

arsinclair:
[...]

PR #53, opened 2024-07-09
feat(Tidal): format textual copyright symbols as © or ℗

phw:
On Tidal it s very common to have copyright symbols in textual form, e.g. "(p)" or "(C)", instead of the dedicated symbols ℗ or ©. The symbols are also used, but it is not mandatory.

Add a utility function `formatCopyrightSymbols` to handle this. Also use this function in the existing Spotify implementation for copyright cleanup.

This PR partially but not completely addresses #23. I wanted to fix the most prominent issue where Tidal shows "(p) 2024 Some Band", but Spotify and iTunes give "℗ 2024 Some Band". Now there is a single consistent copyright entry.

Previously for https://tidal.com/album/64944658:
grafik

Now:
grafik

What this PR does not address:

- Handling copyright entries on Tidal without any symbol at all (textual or not). This needs more consideration, see my comment https://github.com/kellnerd/harmony/issues/23#issuecomment-2157983729
- Considering Tidal copyright entries on track level
- Considering any other provider (but only ones with copyright right now are iTunes and Spotify, and those seem to be fine)


kellnerd (approved):
Thank you, LGTM

PR #62, opened 2024-09-27
feat(web): Set `musicbrainzid` as parameter for MagicISRC

RustyNova016:
Same as https://github.com/kepstin/magicisrc/pull/15.

This prevents false positives on browsers that remove trackers in URLs


kellnerd:
Thank you, the change itself is trivial enough.

When I tested this change, I noticed that the MagicISRC website seems to rewrite the alternative parameter name `musicbrainzid` to the problematic `mbid` immediately after the page loads...
Have you tested whether this actually solves the problem for your browser? I can't test this myself as I never had this issue, but it looks like the updated MagicISRC may only delay the problem.

RustyNova016:
I just tested the live version which is different from my PR, and it does works.

The only bug is after login in MusicBrainz, the redirection also glitches out. I'll try another PR, but this one is good to go

kellnerd (approved):

PR #67, opened 2024-10-31
Use locale date format

c0b41:


kellnerd (changes requested):
Hi, I am sorry, but this won't work for multiple reasons.
The most important one being that we are dealing with partial dates here, for example year and month only:
```ts
new Date(Date.UTC(2024, 10, undefined)).toLocaleDateString() === "Invalid Date"
```

Is there any particular reason you want to change date formatting? The MusicBrainz website uses exactly the same format for (partial) dates.

c0b41:
Sorry, date format was confusing for my locale date format. probably best option use i18n for this stuff. and i see some releaseDate not full date.

PR #77, opened 2024-12-11
Isrc links

### LINKED ISSUE #76: ISRC links on Release Lookup page
Discussed in https://github.com/kellnerd/harmony/discussions/74

<div type='discussions-op-text'>

<sup>Originally posted by **@reubot** December  3, 2024</sup>
I would like to suggest adding both musicbrainz and ifpi links to the ISRC column on the Release Lookup page.
eg
https://isrcsearch.ifpi.org/?tab=%22code%22&isrcCode=%22isrc%22&itemsPerPage=100&showReleases=true
https://musicbrainz.org/isrc/isrc</div>

reubot:
Add links to ISRC search pages for both MB and IFPI (closes #76)

[Beethoven_ Symphonies Nos. 7 & 8 - Harmony.html.gz](https://github.com/user-attachments/files/18091623/Beethoven_.Symphonies.Nos.7.8.-.Harmony.html.gz)

suggested next step is to seed recordings with matching ISRCs for new releases when MB provider is checked


server/components/ISRC.tsx
```diff
@@ -1,16 +1,34 @@
+import { ProviderIcon } from '@/server/components/ProviderIcon.tsx';
+import { join } from 'std/url/join.ts';
+import { SpriteIcon } from '@/server/components/SpriteIcon.tsx';
+import { SpriteIconWidth } from '@/server/components/SpriteIconWidth.tsx';
```
kellnerd, line 4:
```suggestion
import { ProviderIcon } from './ProviderIcon.tsx';
import { SpriteIcon } from './SpriteIcon.tsx';
import { join } from 'std/url/join.ts';
```
Drop now unused import, shorten import specifiers and separate internal and external modules.

kellnerd (changes requested):
Thank you, I like how you tried to keep the ISRC column compact by using icons as links.

Apart from a few nitpicks, I am a bit worried about using the official IFPI logo:
- It is an incomplete version which is hard to recognize, the full version with the "ifpi" text is too detailed for the size used in Harmony.
- It doesn't fit the style of the other stroke-only icons.
- I have no idea about the legal/copyright implications of including the original vectors in this repository.

So what do you think about creating a simple 24x24 grid vector version of their favicon instead? It is just the text IFPI in capital letters which should be easy to design in Inkscape using one of the other custom SVG files as a template.

> suggested next step is to seed recordings with matching ISRCs for new releases

I agree that this would be useful, especially for compilations.
It probably isn't as straightforward as this PR, because we have to be careful not to use recordings where the ISRCs are ambiguous and the user should be able to opt out.
Can you create a ticket to discuss this feature in detail?
Depending on the implementation, it might even make the MB links from this PR useless 😅

server/components/ISRC.tsx
```diff
@@ -1,16 +1,34 @@
...
 
 	return codeMatch
 		? (
-			<code class='isrc'>
-				<span class='country'>{codeMatch[1]}</span>
-				<span class='registrant'>{codeMatch[2]}</span>
-				<span class='year'>{codeMatch[3]}</span>
-				<span class='designation'>{codeMatch[4]}</span>
-			</code>
+			<span>
```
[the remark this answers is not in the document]
reubot: no worries, I haven't used tsx before so wasn't aware of <>

server/icons/BrandIfpi.tsx
```diff
@@ -0,0 +1,28 @@
...
+			class='icon icon-tabler icon-tabler-brand-ifpi'
+			width={size}
+			height={size}
+			viewBox='0 0 24 24'
+			stroke-width={stroke}
+			stroke={color}
+			fill='none'
+			{...props}
+		>
+			<g transform='translate(-343.3 -311.3)' stroke-linecap='butt' stroke-linejoin='butt'>
```
kellnerd, line 19:
```suggestion
			<g transform='translate(-343.3 -311.3)' stroke-linecap='butt' stroke-linejoin='miter'>
```
The old value is not valid for stroke-linejoin, only for linecap. Setting this to the default value will fix the TS error and causes no visible difference.
But there is a general issue with `icon-sprite.svg.tsx` which I will address after merging, you shouldn't be forced to set these default values on the `g` element.

kellnerd (approved):
LGTM, although I am still not happy that there seems to be no better option for the icon.
I would have loved to have one that matches the rest of the icon set, but my own attempts at it did not look as good as yours.

kellnerd:
Just dropping my latest attempt at drawing a Tabler-style IFPI icon here for now:

ifpi-tabler opt

It uses a fixed stroke width of 1px for the four letters, the curve uses the standard stroke of 2px on the 24x24 grid.

Unfortunately this seems to be too small for a browser when rendering them at size 18, at least in Firefox the two i-dots are not visible.
Strangely this rendering issue only happens in context of the Harmony website, the SVG file itself renders just fine in Firefox when I scale it to 70%...

kellnerd:
> Unfortunately this seems to be too small for a browser when rendering them at size 18, at least in Firefox the two i-dots are not visible.
> Strangely this rendering issue only happens in context of the Harmony website, the SVG file itself renders just fine in Firefox when I scale it to 70%...

It turns out I forgot to set the stroke-linecap="round" stroke-linejoin="round" properties in the TSX version 🙈
Now fixed as part of e83178a9b400cd41e5918bd114725ccdf6b3e393, the new IFPI icon is included in today's release.

Issue #65, opened 2024-10-18
Tidal API returns only tracks that are available in the queried region


kellnerd:
[This VA release](https://tidal.com/browse/album/353004537) should have 100 tracks, but Harmony only loads 99 of them, [with track 16 missing](https://harmony.pulsewidth.org.uk/release?tidal=353004537&region=GB&ts=1729267831).

No idea how this can be solved, the first page of returned tracks from the API only contains 99 tracks although we set a limit of 100: https://openapi.tidal.com/albums/353004537/items?countryCode=GB&limit=100&offset=0
I also experimented with higher offsets of 100 and 200 just in case, but they all return the following response: `{"data":[],"metadata":{"total":99}}` (when it should probably say `"total":100` on each page)

ComradeMike:
https://api.deezer.com/track/2717935892

Looks like that track isn't available in some countries, such as the UK. It shows the track for me on the Tidal website, but not in the phone app. Changing it to a country where it is available (eg the US) loads the missing track: https://harmony.pulsewidth.org.uk/release?tidal=353004537&region=US&ts=1729288929

If you want to test with a more obvious case of a release where some tracks have limited availability (and also doesn't hammer MB with API requests):
https://www.qobuz.com/gb-en/album/greatest-hits-catatonia/0809274919366
https://www.qobuz.com/us-en/album/greatest-hits-catatonia/0809274919366
https://harmony.pulsewidth.org.uk/release?spotify=0Oej4mOTCxm9VNZz3xHtKw&gtin=809274919366&deezer=&region=GB&ts=1729288481
https://harmony.pulsewidth.org.uk/release?gtin=809274919366&tidal=&region=GB&ts=1729288482

ComradeMike:
Looking at my above example again, I realised track 10 (the one with 0 available regions) is missing on the Tidal website: https://tidal.com/browse/album/4670631. So the website won't show a track if it's not available somewhere. It even lets me play previews of tracks 1, 4 and 8, despite them not being available in the UK (and claims I can listen to the full track by signing up 🤨).

phw:
At least here locally I get 100 tracks for region GB with the Tidal v2 implementation from #81 . Not sure if this indicates the issue would be solved, though, as I cannot compare to previous behavior.

kellnerd:
> At least here locally I get 100 tracks for region GB with the Tidal v2 implementation from #81

[...]

PR #86, opened 2025-02-20
feat(Tidal): Show a warning if API returns not all tracks for the release

phw:
Show a warning if Tidal returns not all tracks for a release. The `numberOfItems` attribute will still hold the correct track count, but depending on region Tidal might not return all tracks.

See also #65. Not sure if this PR can be considered a full solution, butt it definitely helps to handle this situation.

E.g. if you query https://tidal.com/album/4670631 with region set to GB or US than only 10 of 15 tracks are returned. With region set to DE or FR it returns even only 6 of 15 tracks.

Screenshot 2025-02-20 at 22-00-42 Greatest Hits – Harmony

### DIFF
--- a/providers/Tidal/v2/lookup.ts
+++ b/providers/Tidal/v2/lookup.ts
@@ -125,6 +125,14 @@ export class TidalV2ReleaseLookup extends ReleaseApiLookup<TidalProvider, Single
 			next = content.links.next;
 		}
 
+		const realTrackCount = rawRelease.data.attributes.numberOfItems;
+		if (items.length < realTrackCount) {
+			this.addMessage(
+				`The API returned only ${items.length} of ${realTrackCount} tracks for ${this.lookup.region}, other regions may have more`,
+				'warning',
+			);
+		}
+
 		const result: HarmonyMedium[] = [];
 		let medium: HarmonyMedium = {
 			number: 1,


providers/Tidal/v2/lookup.ts
```diff
@@ -125,6 +125,13 @@ export class TidalV2ReleaseLookup extends ReleaseApiLookup<TidalProvider, Single
 			next = content.links.next;
 		}
 
+		if (items.length < rawRelease.data.attributes.numberOfItems) {
+			this.addMessage(
+				`The API returned only ${items.length} of ${rawRelease.data.attributes.numberOfItems} tracks.`,
```
kellnerd, line 130:
```suggestion
		const realTrackCount = rawRelease.data.attributes.numberOfItems;
		if (items.length < realTrackCount) {
			this.addMessage(
				`The API returned only ${items.length} of ${realTrackCount} tracks for ${this.lookup.region}, other regions may have more`,
```
We can also mention the looked up region and give a hint that trying other regions sometimes helps.
I've dropped the full stop as other messages don't have one either, some are not even full sentences.

kellnerd (approved):
Good idea, I don't think there is anything else we can do.
Looks good to me, I've checked that the count is correct for multi-disc releases and played a bit with the wording of the message.

phw:
@kellnerd Looks good. I squashed this change into the original commit.

Issue #82, opened 2025-01-24
Bandcamp modified dates, a new option and/or a warning?


UltimateRiff:
I don't know if it's revealed in the API, but Bandcamp shows up to at least three dates for albums in the page source, the release date, the publish date, and the modified date. I forget what the difference between the first two is, but the third is probably the most helpful, as it shows when the most recent change to the album was (for example, a tracklist change or an album art change, both of which should get a new MusicBrainz release).

I believe the modified date also includes other modifications which wouldn't likely get their own release (like editing the credits or description), so I don't know that we should automatically use that date for the release date, but having a warning and the option to use that date would be a good addition

an example release: https://vyletpony.bandcamp.com/track/alicorntwilight
and the (current) dates given (through the Bandcamp importer script
![Image](https://github.com/user-attachments/assets/3cf44223-cf22-4142-8ce5-47491089c53b)

phw:
Where did you get the "February 20, 2013" date from? I can't find this anywhere in the source.

In the metadata there are two dates (see https://validator.schema.org/#url=https%3A%2F%2Fvyletpony.bandcamp.com%2Ftrack%2Falicorntwilight), the `datePublished`, which is 2013-02-21 for your example, and the `dateModified`, which is 2024-12-27 15:07:36.

> but the third is probably the most helpful,

The most helpful is the `datePublished`, as it is what usually corresponds to the release date. It's the date on which the release becomes available, i.e. for pre-releases this is also in the future.

The `dateModified` is like an editing timestamp for the entire entry. Whenever anything is edited the timestamp changes. This would also include fixing typos, purchase options or whatever.


> so I don't know that we should automatically use that date for the release date

Definitely not, as it would be inaccurate in most cases.

> but having a warning and the option to use that date would be a good addition

We should be careful how to phrase such warning. I understand the idea: If the release on BC got edited in a way significant enough to warrant a new release, then have the modification date being used as release date. The problem with this is:

1. It is impossible to tell, whether any release significant change has happened to the page.
2. It is impossible to tell the date of such change.

For you above example: Without further information from the time of release we can't know whether any significant change happened. If we assume there was one, it could have been any day between February 21, 2013, and December 27, 2024.

I'd say the likelihood of December 27, 2024 being incorrect as the release date is higher than February 21, 2013 being correct.

UltimateRiff:
[...]

PR #88, opened 2025-02-21
feat(Bandcamp): Use publish date for releases before Bandcamp's launch

phw:
If the Bandcamp release date is from before the launch of Bandcamp in September 2008 than use the separately provided publish date instead.

See #82. This matches the behavior of the Bandcamp importer user script, see https://github.com/murdos/musicbrainz-userscripts/blob/master/bandcamp_importer.user.js#L60-L73


kellnerd (approved):
Looks good to me.
One day we will hopefully make use of the `launchDate` property for every provider using a generic solution which shows warnings and/or discards implausible dates.

PR #89, opened 2025-02-26
parseGermanDate() fixed that it also finds “March”

Eichi76:
\w does not capture German umlauts. Unicode for ä added to the pattern so that 'März' is also found.


kellnerd:
Good catch, looks like I should add a few tests for this function 😅
I will manually pick your change because GitHub would merge this on top of my temporary "WIP" commit.

phw:
@kellnerd Was it intentional this commit is not on the main branch?

kellnerd:
@phw Yes, this is intentional. The method `parseGermanDate` (and the code which makes use of it) are only available on this feature branch so far. It is for a new provider which is still unfinished.

PR #93, opened 2025-03-16
Setup a framework for provider tests

### LINKED ISSUE #83: Store `HarmonyRelease` URL properties as `string` instead of `URL`
I am considering to abandon using the `URL` type for properties of [`HarmonyRelease`](https://github.com/kellnerd/harmony/blob/d51386f27c1af2dd3a493ba6f75ff0278e0667fa/harmonizer/types.ts#L43) in favor of plain strings, which are easier to handle in a number of scenarios:

- There can be serialization/deserialization issues with dynamic Fresh components:  An `URL` property may silently turn into `string` properties when they are rendered on the client-side
- It becomes easier to have a JSON API later when everything is JSON already (which `URL` is not)
- No more [issues](https://github.com/kellnerd/harmony/commit/753acac52206526cd82bf2a02704008b6cdf7aa6) when using `structuredClone` during merging
- Smaller test snapshots without [redundant URL serializations](https://github.com/kellnerd/harmony/commit/4506b1f7380ea3aa6eae5beceb3129ce7adf379f)

The main disadvantes which I see are:

- Less type safety: A string is not guaranteed to be a valid URL
- There are cases where we have to construct an `URL` from the string again to use [e.g. the domain](https://github.com/kellnerd/harmony/blob/4506b1f7380ea3aa6eae5beceb3129ce7adf379f/server/components/Release.tsx#L85)
- It is a **breaking change** which requires changes to all providers and test snapshots

This should affect external links, API links and image URLs only.

Please let me know if you see any other potential downsides or have alternative ideas.

kellnerd:
I am opening this PR for two reasons:
1. I want to verify that the new tests also run on GitHub Actions.
2. I've spent a long time trying to get these provider tests right and would appreciate a second opinion:
   - Is the documentation sufficient for somebody else to run these tests?
   - Is it simple to understand the test specification object of a provider? If not by the code alone, do the JSDoc comments/tooltips of the properties help?
   - After answering the previous questions: Do you feel able to quickly write an URL specification and/or a release lookup specification for another provider? (That would be for Beatport, MusicBrainz, Tidal v2 videos or iTunes.)

Also closes #83.


derat:
Since you asked in the meeting: I glanced through this and the test structure makes sense to me (without any prior experience looking at the code). :-)

mwiencek:
I was able to follow the instructions and run the tests fine on macOS (after upgrading Deno to 2.x and removing the lockfile), and the helpers you've added in providers/test_stubs.ts look...helpful! 🙂 I can't confidently say how quickly I'd be able to write such tests without refreshing my memory of the codebase, but it looks very clean and well-organized overall.

kellnerd:
Thank you for the feedback @derat and @mwiencek!

I wonder whether you got the [same error as GH Actions](https://github.com/kellnerd/harmony/actions/runs/13885230407/job/38849047385) when it was still running Deno 1.46.3.
To be honest, the error message there makes a lot of sense actually, but Deno 2.x versions no longer seem to complain about this, so I simply updated Deno in CI to get it working 😅

mwiencek:
Yep, well first I got an error about `deno test -RE providers/` needing `-RE` to go after `--`. So after running `deno test -- -RE providers/` I then got the `fetch` error you linked. Then after realizing you also upgraded Deno here, I did the same and got an error about mismatched hashes in the lockfile. Deno's fun! 😆

Issue #95, opened 2025-03-21
seed Bandcamp relationship types for recording URLs


UltimateRiff:
related to #37, it'd be great to seed relationship types for Bandcamp URLs (stream for free, download for free, and purchase for download), if the Bandcamp API allows for it.

if you need some examples:

- https://0101.bandcamp.com/track/snailchan-adventure (stream but no purchase or download)
- https://0101.bandcamp.com/track/mal-damour (stream, purchase, and download)
- https://the-heavy.bandcamp.com/track/oh-no-not-you-again (no stream, purchase, or download, URL should be excluded since there's no relationships to be added)
- (there might be other combinations, not sure)

UltimateRiff:
extra example, tracks can be paid only even when the release is free or paid, like https://painesvillesquarerecords.bandcamp.com/track/no-game

kellnerd:
For the **other providers** it is much easier to fill the link types automatically (for recordings as well as artists and labels), because there is usually only one option available (for Deezer, Spotify, Tidal) or one option which is always correct (for Apple Music, Beatport).
So strictly speaking, the user still has to check whether Apple Music / Beatport links should have the download type in addition to the streaming type, but there isn't really a meaningful way to do this of which I'm aware.

**Bandcamp** on the other hand has its own special link type "bandcamp" for artists and labels, while it offers the three choices for recordings and releases. While it is technically possible to fill these correctly, it would require doing one additional request per track for accurate results.
Since Bandcamp doesn't provide an API, this would mean scraping more HTML pages which is very unlikely to become a default behavior as it would significantly slow down the release actions page.

Maybe we could have it behind an option, but I'm not convinced this is worth it just for the link types, the only additional data which we can only get from track pages are ISRCs (which would be useful if there is no other provider) and stuff like lyrics or download links which we shouldn't use.

UltimateRiff:
if we have to rely on HTML scraping, we could prolly just scrape on the release page then? at least for stream for free, possibly for the other two too. on the The Heavy example, it only has play buttons for the streamable tracks (at least on desktop, mobile does have different formatting, tho I think it's still the same HTML page)

https://the-heavy.bandcamp.com/album/the-house-that-dirt-built

there's also buy links that show up on a hover, I'm away from my PC right now, so I don't know if similar is true for free download only releases like this one: https://papercranecollective.bandcamp.com/album/paper-crane-vol-1

[...]

PR #118, opened 2025-06-07
feat: populate track relationship types for Bandcamp

arsinclair:
- stream for free, download for free and purchase for download are populated correctly;
- pay what you want is not populated, because that requires some redesign (to query the individual tracks from bandcamp and check their pricing options); more info here #95.

### DIFF
--- a/providers/Bandcamp/mod.ts
+++ b/providers/Bandcamp/mod.ts
@@ -378,21 +378,18 @@ export class BandcampReleaseLookup extends ReleaseLookup<BandcampProvider, Relea
 			title = title.replace(`${artist} - `, '');
 		}
 
-		const externalIds = title_link ? this.provider.makeExternalIdsFromUrl(title_link, this.rawReleaseUrl) : [];
-		if (externalIds.length > 0) {
-			const linkTypes = this.provider.getTrackLinkTypes(rawTrack);
-			if (linkTypes.length > 0) {
-				externalIds[0].linkTypes = linkTypes;
-			}
-		}
-
 		return {
 			number: trackNumber,
 			title,
 			artists: artist ? [this.makeArtistCreditName(artist)] : undefined,
 			length: rawTrack.duration * 1000,
 			recording: {
-				externalIds,
+				externalIds: title_link
+					? this.provider.makeExternalIdsFromUrl(title_link, {
+						baseUrl: this.rawReleaseUrl,
+						linkTypes: this.provider.getTrackLinkTypes(rawTrack),
+					})
+					: [],
 			},
 		};
 	}
--- a/providers/Bandcamp/mod.ts
+++ b/providers/Bandcamp/mod.ts
@@ -1,5 +1,5 @@
 import { DownloadPreference } from './json_types.ts';
-import type { AlbumCurrent, PlayerData, PlayerTrack, ReleasePage, TrackCurrent, TrackInfo } from './json_types.ts';
+import type { AlbumCurrent, PlayerData, ReleasePage, TrackCurrent, TrackInfo } from './json_types.ts';
 import type {
 	ArtistCreditName,
 	Artwork,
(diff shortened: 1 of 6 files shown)


providers/Bandcamp/mod.ts
```diff
@@ -109,11 +109,47 @@ export default class BandcampProvider extends MetadataProvider {
...
+		const linkTypes: LinkType[] = [];
+
+		if ('track_streaming' in track) {
+			const playerTrack = track as PlayerTrack;
+			if (playerTrack.track_streaming) {
+				linkTypes.push('free streaming');
+			}
+
+			return linkTypes;
+		}
```
kellnerd: It looks like the tracks from the player data have no attribute which identifies them as downloadable.
We could fallback to the `TrackInfo` object for that, since we already have that in those cases where we load the embedded player data.

Since we only need the track lengths from the `PlayerTrack` for `convertRawTrack`, it is probably more elegant to copy them into the `TrackInfo` object. This would allow us to simplify `convertRawTrack` and `getTrackLinkTypes` to accept only `TrackInfo` objects.

providers/Bandcamp/mod.ts
```diff
@@ -349,13 +385,21 @@ export class BandcampReleaseLookup extends ReleaseLookup<BandcampProvider, Relea
 			trackNumber = rawTrack.tracknum + 1;
 		}
 
+		const externalIds = title_link ? this.provider.makeExternalIdsFromUrl(title_link, this.rawReleaseUrl) : [];
+		if (externalIds.length > 0) {
+			const linkTypes = this.provider.getTrackLinkTypes(rawTrack);
+			if (linkTypes.length > 0) {
+				externalIds[0].linkTypes = linkTypes;
+			}
+		}
```
kellnerd: Since we probably need the same logic to inject the link types into the one external ID for the iTunes provider as well, this could be moved into `makeExternalIdsFromUrl`.
That function should then combine `baseUrl` and `linkTypes` into an options object parameter:
```ts
makeExternalIdsFromUrl(entityUrl: URL | string, options: {
	baseUrl?: URL | string;
	linkTypes?: LinkType[];
} = {}): ExternalEntityId[] {
```

providers/Bandcamp/mod.ts
arsinclair: Good suggestion, added it.

The `PlayerTrack` interface is no longer used anywhere outside of the types file, so technically doesn't need to be exported, but I kept the export anyway, in case we will use it in the future.

providers/Bandcamp/mod.ts
arsinclair: I thought about this at first, but then realised that we will need to pass `TrackInfo` to `makeExternalIdsFromUrl` function for this purpose, however _TrackInfo_ is a concept, that exists in the Bandcamp importer only, so it wouldn't work for other importers that may override that function later.

arsinclair: Even if we try to override this function in the Bandcamp importer file with the custom implementation, we still need to declare a correct interface in the base.ts file, making it context-dependent, which I think is not good.

The fact that iTunes might need similar logic doesn't mean it should be moved to the base class - instead, iTunes should implement its own similar pattern in its provider-specific code.

kellnerd: Sorry if I wasn't clear, I didn't intend to move the whole Bandcamp-specific code into `makeExternalIdsFromUrl`, just the reusable part. The link type computation still remains unique to each provider. Since this feels easier to code than to explain, I've pushed a commit with the change 😅

kellnerd (approved):
Looks good to me, we can merge this as-is and leave the bigger changes to support pay-what-you-want tracks for a separate PR.
Or is there anything which you would like to add, more lookup test cases for example?

providers/Bandcamp/mod.ts
arsinclair: Thanks, I see what you meant now! Thanks for updating the PR, the change makes sense to me.

arsinclair:
Looks good! I probably won't have time to implement the PWYW logic this weekend as it requires a much bigger churn, so let's merge this as is (at least it saves some trouble) and I'll try to have the other part ready during the next weekend.

PR #123, opened 2025-06-25
Youtube Music provider

feathecutie:
This PR provides an initial draft of a YouTube Music provider for harmony. I intend to continue working on this, but I wanted to try to get some eyes on this already, both for potentially some feedback and help with things I'm unsure about, and also as a general motivator for me to keep working on this.

What works well (enough) already:
- Resolving GTINs to YouTube Music releases
- Looking up the URLs of YouTube Music releases and extracting the most essential info from them (but sadly, no GTIN, as I don't think YouTube provides this anywhere :/)
- Extract tracks from releases and provide the most essential info
- Extract artists (at least for the release, not for individual tracks yet)
- Amazingly enough, the data we're looking up here does not even need any authentication/token/cookies (at least it works well in my testing without any, but that might be a fluke since I was using cookies until very recently. In case the requests do fail while testing, see [this section](https://ytjs.dev/guide/authentication.html#cookies) in the YouTube.js guide for how to get authentication cookies)

Things I'm not yet happy with / still unsure about:
- As mentioned before, there seems to be no way to extract the GTIN from YouTube Music releases. This means that YouTube Music is not a great initial data source for a release that has no other providers yet, especially since YouTube does not expose nearly as much metadata as other providers do. If anyone does know of a way to get GTINs from YouTube Music albums, please let me know^^
- This draft currently uses the [YouTube.js](https://github.com/LuanRT/YouTube.js) library in order to communicate with YouTube's internal API. I've noticed that harmony currently has no other provider-specific dependencies, and I'd understand if it isn't ideal to pull in such an extensive dependency for a single provider.
  One *could* in theory try to reimplement the parts of YouTube.js we are using, but the internal API that YouTube.js is using is hardly documented and most likely very sensitive to slightly incorrect requests that YouTube.js already correctly handles, not to mention that keeping up with API changes seems unnecessarily hard.
- As mentioned in the previous point, YouTube.js uses the internal YouTube API, also known as "InnerTube". This is sadly necessary as YouTube music provides no consumer-facing API on it's own, and the normal YouTube API is not powerful enough for the data we're looking up. While I haven't encountered any issues and have heard of legal consequences regarding this, I'm not even close to a lawyer, and this might potentially be questionable from a legal standpoint. I'd love some feedback on whether the usage of this kind of API is okay in harmony, because I really don't think a YouTube Music provider is possible without it :(
- I'm not familiar with how often parts of the internal YouTube API actually change, but this provider might definitely be a bit higher maintenance than other providers, although YouTube.js does luckily take care of all the messy internals regarding this
- And as a final pain point.. my code for fetching data is just really messy at some points, and requires more API calls than I'd like. Especially the part for converting between playlists and albums (starting at line 122 in `YoutubeMusic/mod.ts`) currently has to use some weird workarounds, but I'm still trying to find a better and more direct way to get the data we need here, and for now, while less elegant it still works

..also, I really hope this paragraph doesn't sound too LLM-generated, this definitely came out a lot more "formal" than I planned^^


providers/YouTubeMusic/mod.ts
```diff
@@ -0,0 +1,265 @@
...
+// See https://ytjs.dev/guide/ and https://ytjs.dev/api/
+
+// Avoid typos when repeating entity types
+const CHANNEL = 'channel';
+const PLAYLIST = 'playlist';
+const ALBUM = 'album';
+const TRACK = 'track';
+
+export default class YoutubeMusicProvider extends MetadataProvider {
+	override entityTypeMap: Record<HarmonyEntityType, string | string[]> = {
```
kellnerd:
```suggestion
	readonly entityTypeMap: Record<HarmonyEntityType, string | string[]> = {
```
Nitpick: There are a couple more irrelevant `override` keywords which should be removed, they are only helpful to indicate that we overwrite non-abstract props of the base class.

providers/YouTubeMusic/mod.ts
```diff
@@ -0,0 +1,265 @@
...
+		}
+	}
+
+	override constructUrl(entity: EntityId): URL {
+		return new URL(
+			(entity.type === PLAYLIST)
+				? `playlist?list=${entity.id}`
+				: (entity.type === TRACK)
+				? `watch?v=${entity.id}`
+				: `${entity.type}/${entity.id}`,
```
kellnerd:
```suggestion
				: `${entity.type === ALBUM ? 'browse' : entity.type}/${entity.id}`,
```
We have to map 'album' back to 'browse' here in order to obtain valid album URLs.
It would probably make sense to use either a switch statement or a mapping object, now that there are four different cases which make the ternary operator rather unreadable 😀

providers/YouTubeMusic/mod.ts
```diff
@@ -0,0 +1,265 @@
...
+			return {
+				type,
+				id: splitId.join(':'),
+			};
+		} else {
+			return {
+				type: 'artist',
+				id,
+			};
+		}
```
kellnerd: It shouldn't be necessary to encode the type into the serialized IDs, they can be distinguished by their prefixes:
- Album IDs seem to have a `MPREb_` prefix (https://tickets.metabrainz.org/browse/MBS-10226)
- Album playlist IDs have the format `OLAK5uy_[klmn][A-Za-z0-9_-]{32}` (https://wiki.archiveteam.org/index.php/YouTube/Technical_details#Playlists)
```suggestion
		if (entityType === 'release') {
			return {
				type: id.startsWith('OLAK5uy_') ? PLAYLIST : ALBUM,
				id,
			};
		} else {
			return super.parseProviderId(id, entityType);
		}
```
We probably should also limit the allowed IDs to these two prefixes as we don't want to support regular playlists with other prefixes.

Maybe it is also safer to invert the condition to avoid mistakenly treating other playlists as albums?
```suggestion
		if (entityType === 'release') {
			return {
				type: id.startsWith('MPREb_') ? ALBUM : PLAYLIST,
				id,
			};
		} else {
			return super.parseProviderId(id, entityType);
		}
```

providers/YouTubeMusic/mod.ts
```diff
@@ -0,0 +1,265 @@
...
+	innertube: Innertube | undefined;
+
+	override constructReleaseApiUrl(): URL | undefined {
+		return undefined;
+	}
+
+	protected override async getRawRelease(): Promise<YTMusic.Album> {
+		// Innertube needs an async context to be created, so create it in the first async context available
+		if (!this.innertube) {
+			this.innertube = await Innertube.create();
```
kellnerd: Innertube performs a couple of initialization requests before it actually starts to lookup the release. These can be cached (in the temp directory): 
```suggestion
			this.innertube = await Innertube.create({ cache: new UniversalCache(false) });
```

Below is the debugging code which I used to log all InnerTube requests (and record their responses):
```suggestion
			this.innertube = await Innertube.create({
				cache: new UniversalCache(true), // store in working dir
				fetch: async (input, init) => {
					const url = new URL(input instanceof Request ? input.url : input);
					console.debug(
						'YouTube.js fetch',
						url.href,
						input instanceof Request && await input.clone().text(),
						// init,
					);
					const response = await fetch(input, init);
					// const path = await urlToFilePath(url, { baseDir: 'testdata' });
					// await saveResponse(response.clone(), path);
					return response;
				},
			});
```
Example output without caching:
```
YouTube.js fetch https://www.youtube.com/sw.js_data
YouTube.js fetch https://www.youtube.com/youtubei/v1/config?prettyPrint=false
YouTube.js fetch https://www.youtube.com/iframe_api
YouTube.js fetch https://www.youtube.com/s/player/e12fbea4/player_ias.vflset/en_US/base.js
YouTube.js fetch https://www.youtube.com/youtubei/v1/browse?prettyPrint=false&alt=json {"browseId":"MPREb_awgUw0NYaZA","client":"YTMUSIC","isAudioOnly":true}
```
Three out of the four init requests disappear when we use the cache (which then contains session data and player data). Only the request for `iframe_api` is repeated for each lookup, likely to check whether the player hash has changed and new player data has to be downloaded.
We can probably get rid of that request as well if we initialize InnerTube only once for each provider instance and not for each release lookup instance.
(Sidenote, mainly for myself: We have to see if it is still possible to potentially change the region/locale per request.)

providers/YouTubeMusic/mod.ts
```diff
@@ -0,0 +1,265 @@
...
+
+		let number;
+		if (item.index?.text) {
+			try {
+				number = parseInt(item.index.text);
+			} catch (_e) {
+				// Leave number undefined if parsing failed
+			}
+		}
+
```
kellnerd: You don't have to parse the track number, Harmony's typings accept `string | number | undefined`.
```suggestion
```

providers/YouTubeMusic/mod.ts
```diff
@@ -0,0 +1,265 @@
...
+				title: item.title!,
+				externalIds: [{
+					type: TRACK,
+					id: videoId,
+					provider: this.provider.internalName,
+				}],
+			},
+			length,
+			number,
+		} as HarmonyTrack;
```
kellnerd: Using `as` in TypeScript is not safe, `satisfies` is probably what you want here, it helped me to find a couple of mistakes:
```suggestion
			type: 'audio',
			recording: {
				externalIds: [{
					type: TRACK,
					id: videoId,
					provider: this.provider.internalName,
				}],
			},
			length,
			number: item.index?.text,
		} satisfies HarmonyTrack;
```
Or just annotate the return type of the method for the same effect, that's even better IMO 😉

kellnerd (changes requested):
Thank you for tackling this! Before I get into any implementation details, I wanted to share a few general thoughts.

There were two things I was worried about regarding the YouTube.js library:

1. **Code bloat**:
   - I ran `deno info lookup.ts | head` and temporarily commented out the `YouTubeMusicProvider` import in `providers/mod.ts` to get a feeling for the impact.
   - without YouTube Music provider: 2.12MB, 225 unique dependencies
   - with YouTube Music provider: 4.13MB, 982 unique dependencies
   - This looks like a lot, but if I include the whole web app with `deno info server/main.ts | head`, it shouldn't matter too much: 14.92MB, 1398 unique dependencies (YouTube included).
   - I've also tried to replicate the relevant requests by hand with `fetch` and failed to make them work, the API expects a couple of headers and a ton of cruft inside the POST body to even consider the request. So we should really use the library and spare us the extra hassle.
2. **Permalinks and caching won't work**: Lookups for all other providers are cached and used for 24 hours, older snapshots can be accessed using permalinks which contain a timestamp.
   - We can't use the usual provider caching mechanism (`SnapStorage`) because we don't have control over individual requests when we use the library.
   - While it is possible to pass a custom `fetch` function to Innertube, we can't proxy the requests to `fetchSnapshot`/`fetchJSON` (which use Harmony's cache) because we would have to pass the timestamp  well.
   - Additionally, as opposed to any proper public API, the URL is always `https://www.youtube.com/youtubei/v1/browse?prettyPrint=false&alt=json` for every lookup, because the relevant lookup parameters are only contained in the body of the POST request. This is not currently supported by Harmony's cache which uses the URL as key.
   - Even if the cache would support distinguishing POST requests by body data and if we hacked Innertube to pass the timestamp along, the API response contains a lot of unnecessary clutter (e.g. describing the player UI), so I wouldn't want to cache the whole thing anyway.
   - I guess our best way forward is to omit YouTube caching for now or to implement separate caching logic which caches the results of the Innertube calls. If you want to rather focus on the lookup implementation, I can help you with this task.

P.S. I know that there are still some open questions and I guess I also have more feedback, but this should be enough to get you started (and it's getting too hot to think here 😅).

P.P.S. One last thing: Please follow https://www.conventionalcommits.org/en/v1.0.0/ (see the existing commits, the scope would be "YouTube" for all follow-ups). Eventually I will use this to generate changelogs for the website 😇

rinsuki:
As a hobby YTM (=YouTube Music) internals researcher, I wouldn't recommend you to automatically scrapes YTM, because...

* current code treats every tracks are free streaming, but actually some tracks requires (YouTube/YTM) Premium subscription which should be mapped as a (not free) "streaming page" in MB.
  * you might be think "then map unavailable track to paid streaming track"... but its not actually true:
  * there are actual "unavailable" track (can't be played even with premium subscription, like some "purchase only" tracks in Apple Music), but both are marked as "unavailable" if you visiting as free/anonymous user.
  * which means, If you wanted to detect free/paid streaming correctly, needs premium account's session.
  * well I think "scrapes using premium account session" is bad idea.
* If you are visiting music.youtube.com/playlist (album page, incl. InnerTube) as free/anonymous user, some tracks are silently replaced with "Music Video" (actual MV, not the "Provided to YouTube by ...." one).
  * this happens even "Provided to YouTube by ..." video is available for free user.
  * workaround: visit www.youtube.com/playlist page, they wouldn't replace tracks with Music Video.
  * but there are new bad news: some albums are broken and wouldn't returns *any* (including free one) tracks in www.youtube.com/playlist page.
  * so yet another workaround comes in: use the YouTube Data API (public one) to get playlist content, create new playlist, and check it from InnerTube.
  * but I think automatically creates playlist sounds suspicious and I think your account might be banned if you do that.
* Harmony have a region selector, but by obvious reasons, InnerTube always uses your IP address to decide region, so region selector will not work.
  * which means some album (e.g. JP exclusive release) will not able to seed.
  * even worse, some albums have a different Playlist/Track ID, based on your region.
* as you know, search `"barcode"` in YTM would returns album which tied to that barcode, yeah? sometimes not.
  * If album have a multiple versions (e.g. early release, different artwork (pretty common in anime songs), different B-side track), YTM sometimes return "other versions" album in the search result.
  * You can see original album in their albums page's "other versions" section, but... we need to add additional GUI in the Harmony i think.
* I couldn't remember, but there are some additional weird behaviour which isn't noted here.

feathecutie:
First of all, thanks a lot for the review @kellnerd!
I applied most of your suggestions, but I definitely still have a lot of work to do here^^
Note that the current commit history of this PR might be kind of weird because I ended up vendoring some dependencies until some issues are fixed, but this is definitely temporary, see the commit message of https://github.com/kellnerd/harmony/commit/ebcf9e4fe6b74c45585513779340742ab20afe48 for details

@rinsuki, thanks for your thoughts!
I know that scraping the interal YouTube (Music) API is definitely not gonna be a clean and easy process, but I'm not sure I agree that using that data is a bad idea in general.

I will definitely try to improve my code to make as few assumptions as possible, as some of the ones it's currently making are clearly wrong (e.g. the "free streaming" thing). I get that these things are really hard to be sure about (e.g. to decide whether a track is paid or unavailable), but in cases where I can be sure about the data, I feel like it is still more valuable to return some correct data than none at all.

Something I am pretty unsure about is the region problem, as I don't think there is a good way to work around this besides using proxies (which would probably cost money to use, and are probably already blocked by YouTube anyways). Innertube does provide a [location parameter](https://ytjs.dev/guide/getting-started.html#location-string) when constructing the instance, with the description "Geolocation setting", but I doubt that that actually does anything useful (haven't tested it yet).

And the barcode thing also seems really messy.. haven't encountered it myself yet but I trust you that it's a thing. This might definitely mean that a YouTube Music provider is a lot less feasible than I thought, but I'd still love to experiment with it for a while before giving up on this^^
Do you happen to have any links/album ids/barcodes/etc lying around where these issues do occur? It's pretty hard to just randomly stumble upon these, and they would be really helpful for testing. If you have any documentation/write-ups for the issues you mentioned, I'd love some references as well because it's kinda hard to find info on this, but I also already really appreciate your comment!

rinsuki:
There are some examples (NOTE, all of them are checked in Japan IP address (where I live), it might behave differently in another regions):

* Every track needs premium subscription, without MV fallback: https://music.youtube.com/playlist?list=OLAK5uy_lM3xYtf_7at1hw_YJHoXQLE__onCwcE0E
* Some track needs premium subscription, and some tracks will be replaced by MV for free users: https://music.youtube.com/playlist?list=OLAK5uy_mFVoLBZmq4mpgtczlpENgiwBOQZXoUVF4
* www.youtube.com will not return any tracks, but music.youtube.com does: https://www.youtube.com/playlist?list=OLAK5uy_lXxOLO7A0n-BrtVdM28AE_ddvYuQRnI7U
* All tracks requires premium subscription, and some of tracks are not playable (even with premium subscription): https://music.youtube.com/playlist?list=OLAK5uy_nWS9Go5IF5kuhgBnb5DJcGn_VtLJ4SKsE
* Barcode mismatch:
  * This album https://musicbrainz.org/release/82ad1706-45d7-4a86-ba7c-00c5fde01505
  * is available on YTM https://music.youtube.com/playlist?list=OLAK5uy_m_c2orq7P9EWO7nLIWVmT7jbqq-xN-Q0g 
  * but if you tried to search its barcode https://music.youtube.com/search?q=%22195533006299%22
  * YTM would return another version's album https://music.youtube.com/playlist?list=OLAK5uy_mlzC0kPEsWKbXja0vWR_E75D4VZycuf1Q
    * → https://musicbrainz.org/release/020c08fb-c69e-4a8f-aa89-2497dfc265c1

providers/YouTubeMusic/constants.ts
```diff
@@ -0,0 +1,52 @@
...
+	'x-youtube-client-name': '67',
+	'x-youtube-client-version': '1.20250219.01.00',
+};
+
+export const YOUTUBEI_BODY = {
+	isAudioOnly: true,
+	context: {
+		client: {
+			hl: 'en',
+			gl: 'US',
```
rinsuki, line 23: I think we might be want to decide those values based of region parameter

providers/YouTubeMusic/constants.ts
```diff
@@ -0,0 +1,52 @@
...
+			screenDensityFloat: 1,
+			screenHeightPoints: 1440,
+			screenPixelDensity: 1,
+			screenWidthPoints: 2560,
+			clientName: 'WEB_REMIX',
+			clientVersion: '1.20250219.01.00',
+			osName: 'Windows',
+			osVersion: '10.0',
+			userAgent:
+				'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/109.0.0.0 Safari/537.36',
```
rinsuki: probably you want to use USER_AGENT in here?

providers/YouTubeMusic/constants.ts
```diff
@@ -0,0 +1,52 @@
...
+			platform: 'DESKTOP',
+			clientFormFactor: 'UNKNOWN_FORM_FACTOR',
+			userInterfaceTheme: 'USER_INTERFACE_THEME_LIGHT',
+			deviceMake: '',
+			deviceModel: '',
+			browserName: 'Edge Chromium',
+			browserVersion: '109.0.1518.61',
```
rinsuki: (note: if you replaced user agent, you might be want to also replace those values, or remove it since server don't care in many cases)

feathecutie: It seems like I'm actually able to remove all body values besides `clientName` and `clientVersion` and everything still works as expected.
I'll probably do this, as this seems a lot cleaner (the current body was mostly just copied from YouTube.js requests), but I'm a bit afraid that YouTube might get.. suspicious, if the requests don't contain these values?
I have no idea what actually gets detected and logged by YouTube, but I'm afraid that this might cause YouTube to flag and block harmony quicker

rinsuki: we wouldn't able to know about they cares or not :P

well yeah, if you want to reduce chance of getting banned from YT, I think you need to put realistic values, or even scrape HTML (YT or YTM web server is calling API when rendering HTML, and it is included in the returned HTML, for takeover to client JS. at this case you still need to call API manually for some cases, e.g. paging or album search)

providers/YouTubeMusic/constants.ts
feathecutie: Setting the `gl` parameter based on the region would probably be a good idea (though it doesn't seem to affect YouTube's behaviour), but I'm not sure how to select one region to use if multiple regions are provided.
As far as I can tell, other providers solve this by using `queryAllRegions` to query each region separately, but this doesn't make sense in this case since the region parameter doesn't affect anything anyways

rinsuki: I didn't know harmony can set multiple regions...

by the way, my main focus is hl parameter, since YTM (sometimes) have a localized tracklists and will be changed by user's language settings. i didn't check it is depending to this parameters, but I think it will.

(sidenote: but I also think change language parameter by region parameter is good or not? i don't know. Apple Music behave like that, but actually some Apple Music region supports multiple language in one regions and shows based of your device's language settings. e.g. JP region have a japanese and english tracklist)

e.g. https://music.youtube.com/playlist?list=OLAK5uy_nXbgxu51sJwZQz69UwVWVvu-j22_IYvA8 this album have a Japanese tracklist and English tracklist, and if your language settings aren't japanese, it should show english tracklist

I think different tracklist for en-US and en-GB are possibly exists, but I don't know about it actually exists or not.

feathecutie: Yup okay, setting `hl` to `ja` seems to work in order to get localized track lists.
I could try to set this dynamically, but I'm still not sure which language to use when.
I think I've read that the MusicBrainz policy is that release and recording titles should always be submitted in their original localisation (so, japanese localisation for Japanese releases), but that would require knowing what the "original" localisation is.
The problem with this is that for example, I've noticed that some releases and tracks tend to have Japanese localisations even if they aren't originally Japanese (like [this album by the German band Rammstein](https://music.youtube.com/playlist?list=OLAK5uy_nNutR-J9j_crO8GqSKSbEF-fQVGnPVBzE))

kellnerd: When the user specifies multiple preferred regions, they will only be used by providers which support the `queryAllRegions` fallback. But if YouTube doesn't "fail" to return data for a certain unavailable region (in some way), this concept doesn't make sense and the first "supported" region of the preferred regions should be used.
So unless there is a meaningful list of supported regions for YouTube, I would simply use the first region.

Regarding the language / `hl` parameter, I wouldn't try to derive this from the region as these aren't always one-to-one mappings. If we arrive at the conclusion that the language is a useful parameter (for YouTube and potentially other providers), Harmony should allow the user the set a preferred language as well. We can even extract the [language from the release URL](https://harmony.pulsewidth.org.uk/about) for some providers (namely Deezer, Spotify, Beatport, maybe Qobuz in the future), which would override the preference input just like a region from a release URL does (for Apple).

providers/YouTubeMusic/constants.ts
kellnerd: Regarding the copied code, I think you should add a copyright line to all affected files, probably `constants.ts` and `api_types.ts`? I am not a lawyer, but I think something like the following should be sufficient for MIT licensed code:
```ts
// Adapted from / Taken from / Inspired by https://github.com/LuanRT/YouTube.js/tree/v14.0.0
// Copyright (c) 2021 LuanRT. MIT license.
```
Not sure about the wording of the first line, you should know better what is the best description.

providers/YouTubeMusic/mod.ts
```diff
@@ -0,0 +1,612 @@
...
+		}, 0);
+
+		// TODO: WIP code for fetching data from the "View song credits" popup
+		// Could be used to improve artist credits,
+		// and to get the videoId of the actual song (instead of the MV) instead of parsing it from the playlist contents
+		// Using this would however require an additional API call per track
+		const creditsEndpoint = item
+			.musicResponsiveListItemRenderer
+			.menu
+			.menuRenderer
```
kellnerd, line 590: Unless we end up removing this code, we have to ensure that the menu exists:
```suggestion
			?.menuRenderer
```
It is missing for unavailable tracks and currently breaks the conversion of [such a release](https://music.youtube.com/playlist?list=OLAK5uy_mFXBCHz3razvnilUmAVny2l4rfXRP9Nc4).

kellnerd (commented):
Sorry that it took me so long to get back to review this PR, I had significantly less time available than expected this summer.
I agree with your summarized TODO items in the updated PR description and only have a few remarks to add (from my weeks-old notes).

Ideally we should avoid fetching the HTML player just to extract the corresponding browse ID for a playlist.
I still have to compare the two data representations side by side and in depth to see if there is any way to work with just one of them even. The annoyingly UI-centric format of the representation certainly doesn't help with that, that's one of the reasons why I've been putting this off for so long...

I don't know about your available time and plans for this PR, but I would like to see it merged eventually, even if it is not fully ready to be released, since it is a lot of code to review already.
We can always merge an early version of the provider and temporarily disable it by removing the changes to the provider registry for now.

Issue #105, opened 2025-05-10
Auto-select "Soundtrack" secondary type based on release title


Aerozol:
I think in cases where these strings are included in the release title, it's safe to auto-select the "Soundtrack" secondary type:

- "original soundtrack"
- "official soundtrack"
- "original game soundtrack"
- "official game soundtrack"
- "original video game soundtrack"
- "official video game soundtrack"
- "original film soundtrack"
- "official film soundtrack"
- "original motion picture soundtrack"
- "official motion picture soundtrack"
- "original movie soundtrack"
- "official movie soundtrack"
- "soundtrack from the"
- "soundtrack to the"
- "soundtrack zum"
- "original-soundtrack zum"
- "ost"
- "o.s.t"
- "video game soundtrack'
- "film soundtrack"
- "movie soundtrack"
- "series soundtrack"
- ?

---

We could give users an extra nudge that this will be auto-set, e.g.:

![Image](https://github.com/user-attachments/assets/48f5a7e0-0d58-4094-90ae-e43fe6a876f2)

If we add an extra nudge, and feel that it makes it pretty fool-proof, we could also make a leap and simplify the list to "OST", "soundtrack" etc

Ser-Erate:
Having the toggle (or at least some clear notification of it so it may be corrected manually) would certainly be helpful, as there definitely are exceptions, 10CC has an album called "The Original Soundtrack" for example, and Saint Motel have an album and 2 EPs called "The Original Motion Picture Soundtrack", none of which are actually soundtracks to anything.

Aerozol:
There are always exceptions - for instance, sometimes streaming services inadvertently have "single" set when artist intent is clearly "EP", but we still seed the release type. I think the error rate here will be a pretty low percentage.

Issue #130, opened 2025-07-15
Metadata post-processing hooks


kellnerd:
Harmony takes most of the metadata from the providers verbatim and does only little [post-processing](https://github.com/kellnerd/harmony/blob/d70a5ef410abe25f9fb6f2f3b210d02ef936ade5/lookup.ts#L305):
- Language guessing and script detection
- Replace DistroKid placeholder labels with [no label] (new)
- Plus some provider-specific post-processing (like #9)

This leads to common metadata inconsistencies and invalid data being seeded to MusicBrainz, for example:
- Different formatting of extra title information (Spotify uses a hyphen instead of parentheses)
- Featured artists in titles (#39)
- Other irrelevant title prefixes and suffixes
- Labels which are copyright holders rather than imprints ([forum post](https://community.metabrainz.org/t/harmony-music-metadata-aggregator-and-musicbrainz-importer/698641/314?u=kellnerd))

Since there is a never-ending supply of different mistakes, it is near impossible to keep up with new requests to fix them and maintain the code for all kinds of exceptions. Sometimes it is not even possible to fix a mistake automatically without a lot of false positives.

Harmony should stay simple at the core, so it makes more sense to have a **flexible hook system**, where metadata post-processing hooks can be registered.
Such a hook can apply changes to the original metadata, either for a specific provider or for all providers.

It should be possible to define simple **search & replace rules** (with support for regular expressions) for some metadata properties (release title, track titles, artists, labels).
In the simplest case, source and target property are the same (e.g. original title to new title), but in a later iteration it should also be possible to move information (e.g. title to artist).
(We may even have a small, Picard-inspired scripting language for advanced use cases in the end?)

Ideally there should be a few **standard rules** which can be enabled/disabled via the web interface and a way to define **custom rules** as well.
Related rules can be grouped into named rule sets, which hide their implementation details to keep the UI compact (unless the user enters the rule set editing mode).
This proposal still needs to be fleshed out, but I think if we do this right and have a way to share custom hooks, we can streamline many manual corrections which an editor may or may not do.

P.S. This is actually my most wanted feature, writing a dedicated ticket for it was long overdue.

PR #139, opened 2025-09-11
Detect soundtrack based on the release title

gustavkj:
Closes: https://github.com/kellnerd/harmony/issues/105

Implements auto-detection of soundtrack releases based the title based on a wide range of common title patterns. Supports both common title formats in English (which often are used for releases in other languages as well), German, Swedish, and Norwegian.

I've added test cases for a wide variation of soundtracks.

In contrast to the existing patterns for detecting release group types from the title, the pattern does not always capture the type and as such I had to come up with a way of associating the pattern to a type. I think this can probably be useful for when detecting more release group types from titles in the future.

### DIFF
--- a/harmonizer/release_types.ts
+++ b/harmonizer/release_types.ts
@@ -22,7 +22,7 @@ const releaseGroupTypeMatchers: Array<{ type?: ReleaseGroupType; pattern: RegExp
 	{
 		type: 'Soundtrack',
 		pattern:
-			/(?:Original\s|Official\s)(?:(?:(?:Video\s)?Game|Motion Picture|Film|Movie|Television|TV|(?:(?:TV|Television)[\s-]?)?(?:Mini[\s-]?)?Series?|Musical)[\s-])?(?:Soundtrack|Score)/i,
+			/(?:Original|Official)\s(?:(?:(?:Video\s)?Game|Motion Picture|Film|Movie|Television|TV|(?:(?:TV|Television)[\s-]?)?(?:Mini[\s-]?)?Series?|Musical)[\s-])?(?:Soundtrack|Score)/i,
 	},
 	// Common soundtrack title: "Soundtrack from the <Medium>", should also match "Soundtrack from the <Streaming service> <Medium>"
 	{
@@ -61,10 +61,9 @@ export function guessTypesFromTitle(title: string): Set<ReleaseGroupType> {
 			return;
 		}
 		const type = match[1] || matcher.type;
-		if (!type) {
-			return;
+		if (type) {
+			types.add(capitalizeReleaseType(type));
 		}
-		types.add(capitalizeReleaseType(type));
 	});
 	return types;
 }
--- a/harmonizer/release_types.ts
+++ b/harmonizer/release_types.ts
@@ -56,6 +56,11 @@ const releaseGroupTypeMatchers: Array<{ type?: ReleaseGroupType; pattern: RegExp
 export function guessTypesFromTitle(title: string): Set<ReleaseGroupType> {
 	const types = new Set<ReleaseGroupType>();
 	releaseGroupTypeMatchers.forEach((matcher) => {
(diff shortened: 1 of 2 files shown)


kellnerd (commented):
Thank you, that looks promising. I'm not at home right now, so it might take me until next week to review your contributions properly.

harmonizer/release_types.ts
```diff
@@ -11,23 +11,60 @@ export function guessTypesForRelease(release: HarmonyRelease): Iterable<ReleaseG
...
-	/\s- (EP|Single|Live)(?:\s\(.*?\))?$/i,
+	{ pattern: /\s- (EP|Single|Live)(?:\s\(.*?\))?$/i },
 	// Generic "EP" suffix
-	/\s(EP)(?:\s\(.*?\))?$/i,
+	{ pattern: /\s(EP)(?:\s\(.*?\))?$/i },
+	// Common soundtrack title: "Official/Original <Medium> Soundtrack" and "Original Score"
+	{
+		type: 'Soundtrack',
+		pattern:
+			/(?:Original\s|Official\s)(?:(?:(?:Video\s)?Game|Motion Picture|Film|Movie|Television|TV|(?:(?:TV|Television)[\s-]?)?(?:Mini[\s-]?)?Series?|Musical)[\s-])?(?:Soundtrack|Score)/i,
```
kellnerd: Minor simplification:

```suggestion
			/(?:Original|Official)\s(?:(?:(?:Video\s)?Game|Motion Picture|Film|Movie|Television|TV|(?:(?:TV|Television)[\s-]?)?(?:Mini[\s-]?)?Series?|Musical)[\s-])?(?:Soundtrack|Score)/i,
```

I wonder whether we could get away with simplifiying this one pattern to just "Original / Official ... Soundtrack / Score"? At least I can't think of any words to fill the gap to make this not a soundtrack.

harmonizer/release_types.ts
```diff
@@ -11,23 +11,60 @@ export function guessTypesForRelease(release: HarmonyRelease): Iterable<ReleaseG
...
+	releaseGroupTypeMatchers.forEach((matcher) => {
+		const match = title.match(matcher.pattern);
+		if (!match) {
+			return;
 		}
+		const type = match[1] || matcher.type;
+		if (!type) {
+			return;
+		}
+		types.add(capitalizeReleaseType(type));
```
kellnerd:
```suggestion
		if (type) {
			types.add(capitalizeReleaseType(type));
		}
```

kellnerd (commented):
I like the general implementation and the large collection of test cases, well done!

My main worry is the additional complexity (and processing time) that we get with each new pattern since these are executed for every release (among them many non-soundtracks).
Currently we try all patterns without exiting early once we have a match. That mostly makes sense for the previously existing patterns where something can be detected as both an EP and Live, for example.
But for the new soundtrack patterns we no longer have to check the remaining ones once we have a match.
So I see a potential to optimize this function by grouping patterns by release type, but you don't have to do that in this PR.

Especially having specific rules for too many languages could be problematic if we are not careful.
I will accept these for now since you have already implemented them (and they are useful for my own selfish needs 😇), but probably won't accept language-specific rules in the future.
My plan is that such rules can be customized client-side later on, as part of the implementation of #130, which will allow us to have disabled rules.

harmonizer/release_types.ts
gustavkj: > I wonder whether we could get away with simplifiying this one pattern to just "Original / Official ... Soundtrack / Score"?

True 🤔 I was mostly looking at the long list of examples in the issue and ready the comment worrying about false positives. But I agree that it most often will be enough with the pattern you mention. It also would cover all "Original <streaming service> <art medium> soundtrack" formats.

gustavkj: Done ✅

kellnerd (approved):

Issue #116, opened 2025-06-06
Handle subscriber-only releases on Bandcamp


kellnerd:
Example: https://arbee.bandcamp.com/album/des-papiers-ii

This release is currently detected as ("paid download" and) "free streaming", which it surely isn't.
Maybe there are also other JSON properties which don't match the expectations of our type definitions, that remains to be investigated.

(Generally we still need a more diverse collection of test cases to be added to the [test spec](https://github.com/kellnerd/harmony/blob/3fb582df109b59b8f036aadf72aaf76f993c2f6a/providers/Bandcamp/mod.test.ts#L44), preferably with a small JSON footprint since they have to be stored in the `testdata/` folder.)

PR #142, opened 2025-09-15
Handle subscriber-only Bandcamp releases

gustavkj:
Closes: https://github.com/kellnerd/harmony/issues/116

Handle subscriber-only Bandcamp releases, making sure _not_ to mark them as `free streaming`. Also, adding a bunch more test cases for Bandcamp.

It also adds a comment about the case where a track link is lacking a type (if the track itself is not free streaming, or individually downloadable), since it is intentionally left for now until another suitable link type [has been added](https://tickets.metabrainz.org/browse/STYLE-2443).

### DIFF
--- a/providers/Bandcamp/mod.ts
+++ b/providers/Bandcamp/mod.ts
@@ -136,6 +136,11 @@ export default class BandcampProvider extends MetadataProvider {
 			}
 		}
 
+		// Track links without any streaming or download relationship are kept.
+		// It is up to the user to decide what to do.
+		// In the future there might be a separate relationship that can be used.
+		// See https://tickets.metabrainz.org/browse/STYLE-2443
+
 		return linkTypes;
 	}
 


arsinclair:
> I also changed it so that if a release contains tracks that are mostly (half or more) free streaming, the track would be marked as free streaming too (instead of every track).

If I understand this correctly, you're adding some guesswork into the algorithm? I haven't got a chance to look at the code, but why would we need something like this, currently Harmony correctly identifies whether the track is streamable or not for every track.

> Also, an issue I noticed that there are cases where the track link is lacking a type (if the track itself is not free streaming, or individually downloadable). But maybe that's part of this issue: https://github.com/kellnerd/harmony/issues/95

It must be a bug! Currently the only thing we cannot identify at the track level is PWYW (pay what you want). If a track is neither PWYW, nor Streamable or a Free Download, then the link shouldn't be inserted at all, I think.

Sorry, I've been too caught up with the work to actually finish #95. But I hope I'll find time for it sooner or later 😅

gustavkj:
> If I understand this correctly, you're adding some guesswork into the algorithm? I haven't got a chance to look at the code, but why would we need something like this, currently Harmony correctly identifies whether the track is streamable for every track.

I wouldn't call it guess work. As the forum post I cited stated, if the release is "fully or mostly" freely streamable it would be listed as free streaming. To me "mostly" counts as at least half the tracks of the release. But then again, that's not the main focus of this PR, so I can change it back and that can be discussed in an issue instead.

> It must be a bug! Currently the only thing we cannot identify at the track level is PWYW (pay what you want). If a track is neither PWYW, nor Streamable or a Free Download, then the link shouldn't be inserted at all, I think.

As I brought up in #95, there are tracks that can only be acquired by buying/downloading the full album. Let's look at [this test case](https://github.com/kellnerd/harmony/pull/142/files#diff-84a86056f410e049f796d81f6e16a095e248e962eb23601c926359e13d36c0caR261-R293), with e.g. this track <https://hipdozer.bandcamp.com/track/futon-feels>. The track cannot be streamed for free and cannot be downloaded individually. Only by purchasing the album can the track be downloaded or streamed. Currently, when linking external IDs for the track using Harmony it adds the Bandcamp track link without a relationship pre-selected. Since the track only can be acquired through the album, none of the available relationships make sense. 🤔 

So, as discussed in #95, if the link should be inserted, a new relationship type would be required in MB, or it could be excluded from being added. Which is also what I feel makes sense (and was [suggested on the forum](https://community.metabrainz.org/t/style-2443-what-to-do-about-bandcamp-links-that-contain-no-streamable-or-downloadable-content/674689/8)). But then as pointed out in the issue, there could be useful to link to the track page for other reasons as well.

image

image

gustavkj:
I've reverted the logic change related to if a release should be listed as free streaming based on if the tracks are freely streamable. 🙂 

> Sorry, I've been too caught up with the work to actually finish https://github.com/kellnerd/harmony/issues/95. But I hope I'll find time for it sooner or later 😅

No worries, there's no stress, I can related!  🙂

arsinclair:
Ah, I see, thank you for explaining. I guess unless we get a new relationship type for this case in MBS, it's better to avoid inserting empty links, since it creates more work for the user to remove them.

Regarding considering a release streamable based on the number of streamable tracks: we should probably take this conversation somewhere else indeed :) ; but personally I don't agree with this approximation: I think only if 100% of the tracks on the release are streamable, then we can mark the whole release as streamable, this is exactly what the bandcamp userscript currently does.

kellnerd (commented):
Thank you for collecting all these release lookup test cases for Bandcamp! I wanted to have more tests for a long time, because I've always been sweating about these edge cases (and the numerous related JSON properties) whenever I had to touch the link type logic.

I agree with taking out the change of the release's streaming availability logic, that is a separate decision. Personally I'm leaning towards "all tracks have to be available" and hence have implemented it that way. Let's take this back to the MetaBrainz forums and make a poll, if you want.
But it would be useful to keep the corresponding example release as a test case nevertheless. Can you please drop your last commit 8ac5786317730ea54d9fd24aad52f3970d741210 (which removes the test data) and instead simply drop f787a699da6596b42dbdc9f4b8321a5b13d4fa26 (with an interactive rebase)?

Regarding the track links without link types, I think it is still useful to seed them and let the user decide whether these should be removed or are worth keeping with a technically wrong link type. They are probably not that common anyway and if they occur they might lead to more requests for [STYLE-2443](https://tickets.metabrainz.org/browse/STYLE-2443) rather than hiding the problem.

gustavkj:
> But it would be useful to keep the corresponding example release as a test case nevertheless.

I was not too fond of the test case I added (a two track release with one streamable track), which is exactly 50% free streamable. I can still re-add it if you want. Instead I found a release with 7/12 free streamable tracks.

> Regarding the track links without link types, I think it is still useful to seed them and let the user decide whether these should be removed or are worth keeping with a technically wrong link type. They are probably not that common anyway and if they occur they might lead to more requests for [STYLE-2443](https://tickets.metabrainz.org/browse/STYLE-2443) rather than hiding the problem.

Good point 🙂 I added a comment to the code now to add this knowledge there as well: [`08a03e0` (#142)](https://github.com/kellnerd/harmony/pull/142/commits/08a03e08450a1a06efa2d01445bfd6e71c8326cf)


Btw, feel free to use squash merge for my PRs or do you want be to squash my commits before you merge? I tend to only rebase my branches before I mark my PRs as ready for review since it can make it harder to review once someone has started to review a PR. And I also like to show my thought process with separate commits, which then looks very verbose on the main branch if merged without squash merge. 😅  But if you have any preference please let me know 🙂

kellnerd:
Oh, one thing I wanted to mention: You don't have to assert a vast amount of release properties when you are already asserting a snapshot of the whole release. Any change of the converted release, whether intentional or not, will already cause the snapshot assertion to fail.

The only reason why the existing lookup test has additional assertions is to ensure that the most relevant objective of the test case will always be guaranteed, even if the snapshot was to be updated without sufficient attention to the changes.

What I want to say is that you don't have to assert e.g. all the link types in every test case, only those which are relevant for this test case and different from the other test cases.
For example, the "subscriber-only release with Creative Commons license" test should probably only assert the presence of the CC link, everything else should already be covered by the previous test.

In the same spirit we don't have to do snapshot assertions for all lookup tests if we only want to check one specific thing that is different.
But for Bandcamp with all of its edge cases specifically, it makes sense to have multiple full snapshot assertions nevertheless.

> Btw, feel free to use squash merge for my PRs or do you want be to squash my commits before you merge?

My favourite kind of merge is a fast-forward merge :innocent: (i.e. no separate merge commit), followed by a rebase when fast-forwarding is not possible. I usually only do a standard merge commit when I think it is worth preserving the exact commit history with original dates and messages.
Squash merging is only my last resort when a PR's commit history is too messy.

Regarding this PR, I would like to keep the separate commits for the test cases, the new feature and the comment about the empty link types (at least).
I don't particularly like commits which are basically reverted within the same PR, so I would love if you could squash those (especially https://github.com/kellnerd/harmony/pull/142/commits/f787a699da6596b42dbdc9f4b8321a5b13d4fa26).
But I don't mind having separate commits as long as you are able to find a good commit message for them. (Normally I would have squashed https://github.com/kellnerd/harmony/commit/d9946d872aff7f2ebe4cd65c0e0ffd0a4367a380 for that reason.)
For me it doesn't matter too much if you were to force-push during or after the review process as long as this doesn't happen too often.
I hope this helps you to understand my git thought process.

kellnerd (approved):
Awesome, thanks for your patience!

> Tell me if there are still too many, or any specific ones that seem superfluous.

You have found a good compromise already. If I were to suggest more minor changes for test cases, I should have written them myself 😀

PR #147, opened 2025-09-28
Use limited pattern for live track title checks

gustavkj:
Splits out pattern for detection of live tracks and uses it instead of using `guessTypesFromTitle` checks titles against other formats meant for the release title. This should also be better for performance since additional/unnecessary pattern matching is avoided.


kellnerd (approved):
Good catch, I didn't notice that the same patterns were also used for track titles.

Issue #15, opened 2024-06-07
Support release group primary type


phw:
It would be good if providers could set the [primary type](https://musicbrainz.org/doc/Release_Group/Type) and if this would be seeded when submitting to MB.

Not all providers will support this, but it is sometimes possible to at least detect singles and EPs. If in doubt a provider should likely keep this field empty.

Some notes on specific implementations:

1. **iTunes:** No specific support, but the suffixes ` - Single` and ` - EP` seem to be commonly added to singles / EPs. These should be stripped (see #9) and then can be used for seeding the primary type as well. a-tisket does this.
2. **Spotify:** Releases have the field `album_type`, which is one of `album`, `single` or `compilation`. Maybe it is too broad to use the album type (better leave it empty and have the user decide), but single and compilation should be fine to use.
3. **Bandcamp:** At least standalone tracks could be detected as "Single".
4. **Tidal:** *no types specified*
5. Guessing the release type from title might work in many cases. Most of the MB submission user scripts do this, see https://github.com/murdos/musicbrainz-userscripts/blob/master/lib/mbimport.js#L302-L325
6. For any provider making use of the [MusicAlbum](https://schema.org/MusicAlbum) schema there is a [MusicAlbumReleaseType](https://schema.org/MusicAlbumReleaseType). Theoretically this supports the types [AlbumRelease](https://schema.org/AlbumRelease), [BroadcastRelease](https://schema.org/BroadcastRelease), [EPRelease](https://schema.org/EPRelease) and [SingleRelease](https://schema.org/SingleRelease).  But e.g. Bandcamp does not make full use of this and seems to use AlbumRelease generally, except for standalone tracks it uses SingleRelease.

Generally it seems that if specific types, in particular single or EP, are detectable, this could be seeded. In most cases a source type of "album", if given, might be too unspecific and better kept out.

In the release editor the primary type can be seeded using the field `type`.

kellnerd:
Thanks for your detailed thoughts, I agree with almost all of them.
Since we likely also want to use "Compilation" and other secondary types, I would add an array/set property to `HarmonyRelease` which accepts both primary and secondary types and let the MB seeding code deal with the MB specific handling of these.

phw:
"Live" is also something that could be potentially be detected. I haven's seen live tracks explicitly marked as such in APIs, but in many cases you find a suffix like "(live)" on the track or even release titles. If the release has this or if all the tracks have it then it seems safe to set the live type.

phw:
[...]

PR #148, opened 2025-09-29
Detect remix based on the release title

gustavkj:
Part of: https://github.com/kellnerd/harmony/issues/15

Detects if a release is a remix based on the title. Basically, does it contain "remix", "remixes", or "remixed". Initially, I tried to use some very clever regexes. But I realized, in most cases it is probably just enough to check if the title contains one of the words.

### DIFF
--- a/harmonizer/release_types.test.ts
+++ b/harmonizer/release_types.test.ts
@@ -84,6 +84,56 @@ describe('release types', () => {
 				title,
 				new Set(['Soundtrack']),
 			])),
+			// Remix releases
+			...([
+				'Human (Paul Woolford Remix)',
+				'Paper Romance (Purple Disco Machine Remix - Edit)',
+				'Paper Romance (Purple Disco Machine Remix) [Edit]',
+				'Paper Romance (Purple Disco Machine Remix) (Edit)',
+				'Paper Romance (Purple Disco Machine Remix; Edit)',
+				"Stay (Don't Go Away) [feat. Raye] [Nicky Romero Remix]",
+				'Anti‐Hero (Kungs remix extended version)',
+				'Remix',
+				'Anti‐Hero (Remixes)',
+				'The One (feat. Daddy Yankee) [The Remixes]',
+				'The Remixes',
+				'The Remixes - Vol.1',
+				'The Remixes, Pt. 1',
+				'Remixes',
+				'Remixes 81>04',
+				'Never Say Never - The Remixes',
+				'Skin: The Remixes',
+				'The Hills Remixes',
+				'MIDI Kittyy - The Remixes Vol 1',
+				'The Slow Rush B-Sides & Remixes',
+				'Remixed',
+				'Remixed (2003 Remaster)',
+				'Remixed Sides',
+				'Remixed: The Definitive Collection',
+				'The Hits: Remixed',
+				'Remixed & Revisited',
+				'Revived Remixed Revisited',
+				'Welcome To My World (Remixed)',
+				'Mörkrets Narr Remixed',
+			].map((
+				title,
+			): FunctionSpec<typeof guessTypesFromTitle>[number] => [
+				`should detect remix type (${title})`,
+				title,
+				new Set(['Remix']),
+			])),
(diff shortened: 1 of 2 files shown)


harmonizer/release_types.ts
```diff
@@ -18,6 +18,8 @@ const releaseGroupTypeMatchers: Array<{ type?: ReleaseGroupType; pattern: RegExp
 	{ pattern: /\s- (EP|Single|Live)(?:\s\(.*?\))?$/i },
 	// Generic "EP" suffix
 	{ pattern: /\s(EP)(?:\s\(.*?\))?$/i },
+	// Common remix title: "Remixed", "The Remixes", or "<Track name> (<Remixer> remix)". Matching titles starting, ending, or containing Remix, Remixed or Remixes.
+	{ type: 'Remix', pattern: /(?:^Remix(?:e[sd])?$|^Remix(?:e[sd])?\W|\WRemix(?:e[sd])?$|\WRemix(?:e[sd])?\W)/i },
```
kellnerd: We can drastically simplify the expression by matching a word boundary:
```suggestion
	{ pattern: /\b(Remix)(?:e[sd])?\b/i },
```

harmonizer/release_types.test.ts
```diff
@@ -84,6 +84,55 @@ describe('release types', () => {
...
+				'Welcome To My World (Remixed)',
+				'Mörkrets Narr Remixed',
+			].map((
+				title,
+			): FunctionSpec<typeof guessTypesFromTitle>[number] => [
+				`should detect remix type (${title})`,
+				title,
+				new Set(['Remix']),
+			])),
+			// Multiple types
```
kellnerd, line 126: Another test case to demonstrate why we can't simply match on `/Remix/i`:
```suggestion
			['should not treat a premix as remix', 'Wild (premix version)', new Set()],
			// Multiple types
```

harmonizer/release_types.test.ts
```diff
@@ -84,6 +84,55 @@ describe('release types', () => {
...
+				'Revived Remixed Revisited',
+				'Welcome To My World (Remixed)',
+				'Mörkrets Narr Remixed',
+			].map((
+				title,
+			): FunctionSpec<typeof guessTypesFromTitle>[number] => [
+				`should detect remix type (${title})`,
+				title,
+				new Set(['Remix']),
+			])),
```
kellnerd, line 124: There are so many examples of remix (and soundtrack) releases now, that I'm wondering whether we should factor these out of the main array in order to make the test output less verbose.
It probably makes sense to independently register only one (or a few) tests with `it` that contain a loop of assertions.
I would do that outside of this PR, what do yout think?

harmonizer/release_types.ts
gustavkj: Neat! 🌟 TIL, I was not aware of `\b` in regex 👍

harmonizer/release_types.test.ts
gustavkj: If I understand correctly, you mean that we would e.g. have one `it` test case where we test the soundtrack pattern and one for the remix patterns. Correct?

If so, I agree it's getting a bit verbose right now. Although, it is nice when developing to get all title cases that are failing instead of just the first one (since keeping them all as asserts in one `it` would stop at the first failed assert).

But maybe it's possible to write the tests in such a way that it would be easy to switch to a more verbose mode. 🤔 Could also be that we could use the dot test reporter by default to make the test print out less verbose also 😉 But I'm not against refactoring the tests.

kellnerd: > If I understand correctly, you mean that we would e.g. have one it test case where we test the soundtrack pattern and one for the remix patterns. Correct?

Yeah, almost. We could still have multiple `it`, e.g. one for each of the "headings" which are currently comments.

> Although, it is nice when developing to get all title cases that are failing instead of just the first one (since keeping them all as asserts in one `it` would stop at the first failed assert).

You're right, that's one caveat. Maybe we can find a good compromise between both approaches and have an additional level of test nesting, for example.
Let me think about that a bit more, I'm going to merge this as-is.

Issue #96, opened 2025-03-25
feat: open all "link external IDs" in single click


Dr-Blank:
Could be 2 clicks: one for all artists and one for all the tracks such that all the links open in new tabs just by clicking once on this new button.

Aerozol:
YoGo9 has created a userscript that adds a "open all" button, which could be integrated:
https://github.com/YoGo9/Scripts/tree/main (HarmonyOpenAllRecordings)

Issue #51, opened 2024-07-01
Copy to clipboard buttons


kellnerd:
These buttons to quickly copy data [have been requested](https://community.metabrainz.org/t/harmony-music-metadata-aggregator-and-musicbrainz-importer/698641/54?u=kellnerd) for (unavailable) region lists and external links, but they also make sense for other data:
- [x] Region lists
- [x] External links
- [ ] Annotation (or its individual sections: copyright, credits, availability...)
- [x] Tracklist (for the MB track parser)
- [x] Barcode?
- [ ] Catalog number?
- [ ] Permalink?

Aerozol:
The ["afro's Harmony add-ons" userscript](https://github.com/afrocatmusic/userscripts/blob/main/afros-Harmony-Add-Ons.user.js) implements a copy to clipboard function quite elegantly - just clicking on the GTIN square (for instance) in the seeder will copy the contents to the clipboard.

This could probably even be added to every field? Without requiring an icon or anything.

Stuff like copying the whole tracklist, in track-parser format, are not solved by this example, and would probably require a separate icon and backend code.

kellnerd:
@afrocatmusic and @Aerozol: I would love to hear your opinions and use cases.

Regarding button placement: I would like this to be consistent, so placing the buttons inside the header cells will probably not work as soon as a cell/row contains multiple copyable values, for example if I combine all the availability/release event stuff one day per #85.

kellnerd:
Thank you for your thoughts @arsinclair, I tend to agree with them.

Regarding the bevel, that's just the default style of all buttons (which I had to define because of [dark and light mode adjustments](https://github.com/kellnerd/harmony/commit/48384a0d9ae8afe4e04d6e91aa29df019ca584f5)). I am open for any CSS improvements.
Or do you just want to suggest to display the copy buttons in a different style with simpler borders or even without borders?

> Never needed something like this, but let me put my 5 cents anyway :)

I don't really have a use for them either, but I guess there are quite a few people who want to update existing releases with Harmony data and want to improve their manual workflow.
I think many of these may become obsolete once Harmony properly supports that out of the box with #73, but I'm still interested to hear about other use cases.

Aerozol:
Great stuff!

It is useful for some people - I use a userscript ([afro's Harmony Add-Ons](https://github.com/afrocatmusic/userscripts)) that lets me quickly copy the barcode feel - but not critical so I agree that the buttons can be styled to be secondary/non-critical.

Something like:

[...]

PR #150, opened 2025-10-04
feat: Open all add external IDs links

gustavkj:
Closes: https://github.com/kellnerd/harmony/issues/96

Adds a button to open all links to add external IDs for artists, labels, and recordings. Since opening multiple tabs/windows is blocked by default in most browsers, a warning is shown if the tabs were blocked informing the user that they need to allow pop-ups for the site.

The button to open all links is only shown when more than one link is available. Some margin around the group of links is added then there are multiple links and the button is shown. 

While I've styled the button as a link, it is semantically a button. But at least in from my point of view, I think it fits in better by following the same format as all the other sections on the release actions page.

**Side note:** I wrote a test for `getMusicBrainzEditLink` (which is stashed locally for now) but it is failing due to write permission always being required for the `snaps.db` file. It would be possible to refactor the function further and use dependency injection to get around this. But thought I'd bring it up here before doing so. Thought on that, or suggestions around mocking/stubbing in Deno test?

**Side note 2:** I'm new to the islands architecture and Fresh/Preact, so feel free to give feedback if there are anything that should be done differently. Initially, I ran into an issue related to this where I was trying to send the URL objects to the island which did not work since it is not serializable.

### Button is available for both artist and recording links (as well as for labels) (`/release/actions?release_mbid=53bbe568-4b33-4e93-b93c-f3adc4e15b2d`)

image

### Message informing the user after having clicked the button if the browser blocked tabs from opening

image

### No change when there is only one link (`/release/actions?release_mbid=7b49ac48-6c32-4831-a7dd-aba21b075299`)

image

### DIFF
--- a/musicbrainz/edit_link.ts
+++ b/musicbrainz/edit_link.ts
@@ -1,0 +1,68 @@
+import { join } from 'std/url/join.ts';
+import { flatten } from 'utils/object/flatten.js';
+import { musicbrainzTargetServer } from '@/config.ts';
+import { convertLinkType } from '@/musicbrainz/seeding.ts';
+import type { EntityType } from '@kellnerd/musicbrainz/data/entity';
+import type { EntityWithMbid } from '@kellnerd/musicbrainz/api-types';
+import type { ExternalLink, ResolvableEntity } from '@/harmonizer/types.ts';
+import type { ProviderRegistry } from '@/providers/registry.ts';
+
+// TODO: incomplete type, expose a suitable type from @kellnerd/musicbrainz?
+// interface $EntityWithUrlRels extends EntityWithMbid, WithRels<'url-rels'> {}
+// type EntityWithUrlRels = WithIncludes<$EntityWithUrlRels, 'url-rels'>
+export interface EntityWithUrlRels extends EntityWithMbid {
+	relations: Array<{
+		url: {
+			resource: string;
+		};
+	}>;
+}
+
+export function getEditUrlToSeedExternalLinks(
+	{ entity, entityType, sourceEntityUrl, entityCache, providers }: {
+		entity: ResolvableEntity;
+		entityType: EntityType;
+		sourceEntityUrl: URL;
+		entityCache?: EntityWithUrlRels[];
+		/**
+		 * Registry of available metadata providers to construct external URLs and get link types.
+		 * Sent as a parameter to allow easier testing/mocking.
+		 */
+		providers: ProviderRegistry;
+	},
+): URL | null {
+	if (!entity.externalIds?.length || !entity.mbid) return null;
+
+	// Get the entity from the cache and check which links already exist in MB.
+	const mbEntity = entityCache?.find((e) => e.id === entity.mbid);
+	const existingLinks = new Set(mbEntity?.relations.map((urlRel) => urlRel.url.resource));
+
+	// Convert external IDs into links and discard those which already exist.
+	const externalLinks: ExternalLink[] = entity.externalIds.map((externalId) => {
+		const provider = providers.findByName(externalId.provider)!;
+		return {
+			url: provider.constructUrl(externalId).href,
+			types: externalId.linkTypes ?? provider.getLinkTypesForEntity(externalId),
+		};
+	}).filter((link) => !existingLinks.has(link.url));
+
+	if (!externalLinks.length) return null;
+
+	// Construct link to seed the MB entity editor.
+	const mbEditLink = join(musicbrainzTargetServer, entityType, entity.mbid, 'edit');
+	mbEditLink.search = new URLSearchParams(flatten({
+		[`edit-${entityType}`]: {
+			url: externalLinks.flatMap((link) =>
+				link.types?.length
+					? link.types.map((type) => ({
+						text: link.url,
+						link_type_id: convertLinkType(entityType, type, new URL(link.url)),
+					}))
+					: ({ text: link.url })
+			),
+			edit_note: `Matched ${entityType} while importing ${sourceEntityUrl} with Harmony`,
+		},
+	})).toString();
+
+	return mbEditLink;
+}
--- a/server/components/LinkWithMusicBrainz.tsx
+++ b/server/components/LinkWithMusicBrainz.tsx
@@ -1,67 +1,64 @@
 import { LinkedEntity } from './LinkedEntity.tsx';
 import { SpriteIcon } from './SpriteIcon.tsx';
+import { OpenAllLinks } from '@/server/islands/OpenAllLinks.tsx';
 
-import { musicbrainzTargetServer } from '@/config.ts';
-import type { ExternalLink, ResolvableEntity } from '@/harmonizer/types.ts';
-import { convertLinkType } from '@/musicbrainz/seeding.ts';
-import { providers } from '@/providers/mod.ts';
-import type { EntityWithMbid } from '@kellnerd/musicbrainz/api-types';
 import type { EntityType } from '@kellnerd/musicbrainz/data/entity';
-import { join } from 'std/url/join.ts';
-import { flatten } from 'utils/object/flatten.js';
+import { type EntityWithUrlRels, getEditUrlToSeedExternalLinks } from '@/musicbrainz/edit_link.ts';
+import type { ResolvableEntity } from '@/harmonizer/types.ts';
+import { isDefined } from '@/utils/predicate.ts';
+import { providers } from '@/providers/mod.ts';
 
-// TODO: incomplete type, expose a suitable type from @kellnerd/musicbrainz?
-// interface $EntityWithUrlRels extends EntityWithMbid, WithRels<'url-rels'> {}
-// type EntityWithUrlRels = WithIncludes<$EntityWithUrlRels, 'url-rels'>
-export interface EntityWithUrlRels extends EntityWithMbid {
-	relations: Array<{
-		url: {
-			resource: string;
-		};
-	}>;
-}
-
-export function LinkWithMusicBrainz({ entity, entityType, sourceEntityUrl, entityCache }: {
-	entity: ResolvableEntity;
+/**
+ * Renders a list of MusicBrainz edit links to add external links, based on the provided entities and entity type.
+ * If multiple entities are present, a button to open all MusicBrainz edit links at once is shown.
+ */
+export function LinkWithMusicBrainz({ entities, entityType, sourceEntityUrl, entityCache }: {
+	entities: ResolvableEntity[];
 	entityType: EntityType;
 	sourceEntityUrl: URL;
 	entityCache?: EntityWithUrlRels[];
 }) {
-	if (!entity.externalIds?.length || !entity.mbid) return null;
+	// No entities or no source entity URL to link from, nothing to render.
+	if (!sourceEntityUrl) return null;
 
-	// Get the entity from the cache and check which links already exist in MB.
-	const mbEntity = entityCache?.find((e) => e.id === entity.mbid);
-	const existingLinks = new Set(mbEntity?.relations.map((urlRel) => urlRel.url.resource));
+	const entitiesWithMbEditLinks = entities.map((entity) => {
+		const mbEditLink = getEditUrlToSeedExternalLinks({ entity, entityType, sourceEntityUrl, entityCache, providers });
+		if (mbEditLink) {
+			return { entity, mbEditLink };
+		}
+	}).filter(isDefined);
 
-	// Convert external IDs into links and discard those which already exist.
-	const externalLinks: ExternalLink[] = entity.externalIds.map((externalId) => {
-		const provider = providers.findByName(externalId.provider)!;
-		return {
-			url: provider.constructUrl(externalId).href,
-			types: externalId.linkTypes ?? provider.getLinkTypesForEntity(externalId),
-		};
-	}).filter((link) => !existingLinks.has(link.url));
+	if (entitiesWithMbEditLinks.length === 0) return null;
 
-	if (!externalLinks.length) return null;
+	const actions = entitiesWithMbEditLinks.map(({ entity, mbEditLink }) => (
+		<LinkWithMusicBrainzAction
+			mbEditLink={mbEditLink}
+			entity={entity}
+			entityType={entityType}
+		/>
+	));
+	if (actions.length > 1) {
+		return (
+			<div class='action-group'>
+				<OpenAllLinks
+					links={entitiesWithMbEditLinks.map(({ mbEditLink }) => mbEditLink.href)}
+					linkType={entityType}
+				/>
+				{actions}
+			</div>
+		);
+	} else {
+		return actions[0];
+	}
+}
 
-	// Construct link to seed the MB entity editor.
-	const mbEditLink = join(musicbrainzTargetServer, entityType, entity.mbid, 'edit');
-	mbEditLink.search = new URLSearchParams(flatten({
-		[`edit-${entityType}`]: {
-			url: externalLinks.flatMap((link) =>
-				link.types?.length
-					? link.types.map((type) => ({
-						text: link.url,
-						link_type_id: convertLinkType(entityType, type, new URL(link.url)),
-					}))
-					: ({ text: link.url })
-			),
-			edit_note: `Matched ${entityType} while importing ${sourceEntityUrl} with Harmony`,
-		},
-	})).toString();
-
+function LinkWithMusicBrainzAction({ mbEditLink, entity, entityType }: {
+	mbEditLink: URL;
+	entity: ResolvableEntity;
+	entityType: EntityType;
+}) {
 	return (
-		<div class='message'>
+		<div class='action'>
 			<SpriteIcon name='link' />
 			<div>
 				<p>
--- a/server/islands/OpenAllLinks.tsx
+++ b/server/islands/OpenAllLinks.tsx
@@ -1,0 +1,89 @@
+import { SpriteIcon } from '@/server/components/SpriteIcon.tsx';
+import { Button } from '@/server/components/Button.tsx';
+import type { EntityType } from '@kellnerd/musicbrainz/data/entity';
+import { useSignal } from '@preact/signals';
+
+/**
+ * Delay between every pop-up, in milliseconds.
+ */
+const popUpDelay = 300;
+const blockedPopUpMessage = 'Pop-up was blocked';
+
+export function OpenAllLinks({ links, linkType }: {
+	links: string[];
+	linkType: EntityType;
+}) {
+	const showBlockedPopUpWarning = useSignal(false);
+	const numberOfLinks = links.length;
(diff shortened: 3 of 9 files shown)


ligeia:
maybe I missed it, but does this have a buffer of any kind in between link opens? if a user opens all the links for a 40-track release at once, for example, they'll likely violate musicbrainz rate-limiting.

as a high volume editor, my number one issue is always MB rate-limiting.

ligeia:
> Does MusicBrainz limit the web server requests? I've never encountered that, usually only API is rate-limited.

I think all the requests from your IP are counted in rate limiting. None of the link opens will be dropped or show an obvious problem, but if there is are API request happening elsewhere at the same time, those will start to fail if too many requests are sent via web at the same time.

Aerozol:
We can handily preview how it would look as a button, because something very similar is implemented in the ["HarmonyOpenAllRecordings" userscript](https://github.com/YoGo9/Scripts/blob/main/HarmonyOpenAllRecordings.user.js):

image

I honestly don't think I have a preference. The text saying how many tabs will be opened seems the most important.

A lazy preview of a h3 header + button:

image

I don't really like the button next to the header, but a header could be useful if this section keeps getting busier.

This is quite a long way to say that I don't really mind the current implementation in this ticket :D

arsinclair:
I want to iterate on Aerozol's stylistic suggestion and recommend to dim the blue colour at least by half (it's visually distracting), and also maybe to make it shorter the text could say `Open All Links`. I personally don't find a lot of utility in knowing how many tabs are about to open.

Thank you for the great work and can't wait for this to get merged! And yes, to double kellnerd's point, a button is a button, we shouldn't make it pretend to be a link (because some old-school users like myself will click on it using a mouse-wheel button and wil be frustrated that nothing is happening 😁).

gustavkj:
Alright, there are some different opinions. It seems a visual button is the most popular option. 🙂 

> because some old-school users like myself will click on it using a mouse-wheel button and wil be frustrated that nothing is happening

I get your point. Buuut... We could also just bump [Preact](https://github.com/preactjs/preact/releases/tag/10.26.0) to get [`auxclick`](https://developer.mozilla.org/en-US/docs/Web/API/Element/auxclick_event) event handling, then middle clicks would be possible on a button too. 😉 

---

Another thought, it could be an idea to show a confirmation box before opening a lot of tabs. Taking inspiration from [`refined-github`](https://github.com/refined-github/refined-github/blob/211e100de1c7c15cd1b9f8bec15d2f1ff55d204e/source/helpers/open-tabs.ts#L7), with _10 or more_ tabs this confirmation could show up:

image

And here's a new design with the buttons looking like buttons. I've also implemented the delay between the tabs opening.

I've kept things in separate commits for now if we want to go back. But if things looks good, I can squash them.

image

arsinclair:
Thank you @gustavkj,  I'm happy with the design of it :)

server/components/LinkWithMusicBrainz.tsx
```diff
@@ -1,65 +1,65 @@
...
-import type { ExternalLink, ResolvableEntity } from '@/harmonizer/types.ts';
-import { convertLinkType } from '@/musicbrainz/seeding.ts';
-import { providers } from '@/providers/mod.ts';
-import type { EntityWithMbid } from '@kellnerd/musicbrainz/api-types';
 import type { EntityType } from '@kellnerd/musicbrainz/data/entity';
-import { join } from 'std/url/join.ts';
-import { flatten } from 'utils/object/flatten.js';
+import { type EntityWithUrlRels, getMusicBrainzEditLink } from '@/utils/mbLink.ts';
+import type { ResolvableEntity } from '@/harmonizer/types.ts';
+import { OpenAllLinks } from '@/server/islands/OpenAllLinks.tsx';
```
kellnerd: Can you group the island import with the other component imports at the top? I would like to keep JSX/TSX imports separated from the rest.

server/components/LinkWithMusicBrainz.tsx
```diff
@@ -1,65 +1,65 @@
...
+
+/**
+ * Renders a list of MusicBrainz edit links to add external links, based on the provided entities and entity type.
+ * If multiple entities are present, a button to open all MusicBrainz edit links at once is shown.
+ */
+export function LinkWithMusicBrainz({ entities, entityType, sourceEntityUrl, entityCache }: {
+	entities: ResolvableEntity[];
 	entityType: EntityType;
-	sourceEntityUrl: URL;
+	sourceEntityUrl?: URL;
```
kellnerd:
```suggestion
	sourceEntityUrl: URL;
```
I don't want to make a required property optional just because it makes it easier to handle in TS, even if that requires a bit more code where the component is used.

server/components/LinkWithMusicBrainz.tsx
```diff
@@ -1,65 +1,65 @@
...
+	if (!sourceEntityUrl) return null;
 
-	// Get the entity from the cache and check which links already exist in MB.
-	const mbEntity = entityCache?.find((e) => e.id === entity.mbid);
-	const existingLinks = new Set(mbEntity?.relations.map((urlRel) => urlRel.url.resource));
+	const entitiesWithMbEditLinks = entities.map((entity) => ({
+		entity,
+		mbEditLink: getMusicBrainzEditLink({ entity, entityType, sourceEntityUrl, entityCache, providers }),
+	}))
+		.filter(isEntityWithMbEditLink);
```
kellnerd: We can avoid having to define the one-off type `EntityWithMbEditLink` and its type guard function:
```suggestion
	const entitiesWithMbEditLinks = entities.map((entity) => {
		const mbEditLink = getMusicBrainzEditLink({ entity, entityType, sourceEntityUrl, entityCache, providers });
		if (mbEditLink) {
			return { entity, mbEditLink };
		}
	}).filter(isDefined);
```

server/components/LinkWithMusicBrainz.tsx
```diff
@@ -1,65 +1,65 @@
...
+					mbEditLink={mbEditLink}
+					entity={entity}
+					entityType={entityType}
+				/>
+			))}
+		</div>
+	);
+}
 
+function LinkWithMusicBrainzEntry({ mbEditLink, entity, entityType }: {
```
kellnerd:
```suggestion
function LinkWithMusicBrainzAction({ mbEditLink, entity, entityType }: {
```
P.S. Continuing this thought, we could probably also introduce a new CSS class `action` which shares the same styles as a `message`? I was just lazy when I initially designed the release actions page 😅
Alternatively we could keep `message` and have an additional semantic class `action` to aid userscripts/userstyles.

server/components/LinkWithMusicBrainz.tsx
```diff
@@ -1,65 +1,65 @@
...
+			)}
+			{entitiesWithMbEditLinks.map(({ entity, mbEditLink }) => (
+				<LinkWithMusicBrainzEntry
+					mbEditLink={mbEditLink}
+					entity={entity}
+					entityType={entityType}
+				/>
+			))}
+		</div>
+	);
```
kellnerd: I only wanted to omit the unnecessary wrapper div for a single link.... which forced me to refactor the entire thing:
```suggestion
	const actions = entitiesWithMbEditLinks.map(({ entity, mbEditLink }) => (
		<LinkWithMusicBrainzAction
			mbEditLink={mbEditLink}
			entity={entity}
			entityType={entityType}
		/>
	));

	if (actions.length > 1) {
		return (
			<div class='message-group'>
				<OpenAllLinks
					mbEditLinks={entitiesWithMbEditLinks.map(({ mbEditLink }) => mbEditLink.href)}
					entityType={entityType}
				/>
				{actions}
			</div>
		);
	} else {
		return actions[0];
	}
```

server/islands/OpenAllLinks.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+ * Delay between every pop-up, in milliseconds.
+ */
+const popUpDelay = 300;
+const blockedPopUpMessage = 'Pop-up was blocked';
+
+export function OpenAllLinks({ mbEditLinks, entityType }: {
+	mbEditLinks: string[];
+	entityType: EntityType;
```
kellnerd: I wonder whether these properties should have more generic names to make reuse of this component more intuitive:
```suggestion
export function OpenAllLinks({ links, entityType }: {
	links: string[];
	entityType: EntityType;
```
Maybe even rename `entityType` to `linkType`, then we could simply relax its type to string if needed in the future.

server/routes/release/actions.tsx
```diff
@@ -194,20 +194,17 @@ export default defineRoute(async (req, ctx) => {
...
+					entities={allArtists}
+					entityType='artist'
+					sourceEntityUrl={releaseUrl}
+					entityCache={mbArtists}
+				/>
 				{release?.labels && (
 					<LinkWithMusicBrainz
 						entities={release.labels}
 						entityType='label'
-						sourceEntityUrl={releaseUrl!}
```
kellnerd: This type assertion is safe because `release` is only defined when `releaseUrl` is defined as well.

musicbrainz/edit_link.ts
kellnerd, line 1: Please move this module (and its tests) to `musicbrainz/edit_link.ts` where other MB-specific code lives. `utils` is just my dump folder for everything which currently doesn't fit elsewhere.

utils/mbLink.ts
```diff
@@ -0,0 +1,68 @@
...
+// type EntityWithUrlRels = WithIncludes<$EntityWithUrlRels, 'url-rels'>
+export interface EntityWithUrlRels extends EntityWithMbid {
+	relations: Array<{
+		url: {
+			resource: string;
+		};
+	}>;
+}
+
+export function getMusicBrainzEditLink(
```
kellnerd: This function needs a description and/or a better name, since it generates an MB edit link to add external linls specifically. Maybe something like `getEditUrlToAddExternalLinks` or `constructUrlToSeedExternalLinks`?

server/static/harmony.css
```diff
@@ -411,6 +420,28 @@ nav img.icon-logo {
...
+	border: none;
+	color: var(--button-accent-text);
+	font-size: inherit;
+	font-family: inherit;
+}
+
+button.open-all-links:focus, button.open-all-links:hover {
+	background-color: var(--button-accent-focus);
+	cursor: pointer;
+}
```
kellnerd, line 438: Some of these properties seem to be leftovers from restyling links as buttons or have no visible effect and can be removed: cursor, display, align-items, white-space
(font-size is defined twice, the first one should be removed.)

Other styles such as font settings and maybe border could probably applied to buttons in general.
Existing buttons already have a different border-radius, for example.

P.S. I always appreciate UI design advice, the current button style was the best thing I could come up within a reasonable amount of time.

ligeia:
> Where can we find that userscript and is it possible to reproduce more sophisticated tab/window management functionality in a regular website? As far as I am aware that only works with browser extensions (userscript extensions make use of the same browser APIs).

There is SOME control we can do in browser but, e.g., methods in Firefox and Chrome are different. Yes, most of the control with that script right now depends on GM-related functions. I will see if I can make some suggestions but the API for Harmony that I'm working on is a higher priority for me at the moment.

server/components/LinkWithMusicBrainz.tsx
gustavkj: I get what your saying. 👍 Personally, I try to avoid non-null assertions most often as they can hide null or undefined values that should be checked for.

In this case, I would prefer to refactor the actions page to make it clearer what data is expected. Either there is a release with corresponding data or there isn't, and in either case there is an array of potential error messages. I think that could be clearer from both a TS perspective and a coding perspective too. But that would be a later refactoring. 🙂 

The `actions.tsx` file was a bit inconsistent, using non-null assertions in some places while including `releaseUrl` in conditions in some places seeing if it was truthy. Should I use the non-null assertion for all references to `releaseUrl`?

server/components/LinkWithMusicBrainz.tsx
gustavkj: Nice 👍

server/components/LinkWithMusicBrainz.tsx
gustavkj: I've added a class called `action` now. As you say, it makes more sense semantically.

server/static/harmony.css
gustavkj: The `cursor` bit was borrowed from here, so it should probably be removed there too? 🤔 

https://github.com/kellnerd/harmony/blob/7289e99599c8e86594285c0477ac44dec8a9f1e6/server/static/harmony.css#L206-L209


> Other styles such as font settings and maybe border could probably applied to buttons in general.
> Existing buttons already have a different border-radius, for example.

This makes sense. I think it is good to make the browsers use the opinionated styles from the Harmony stylesheet. Right now they use their built-in font sizes for buttons and input elements. If this is just changed for buttons or submit buttons, text fields look a bit small. So, the question is if they should get the same treatment?

**Current design**

image

**Inherited font-sizes for only buttons**

image

**Inherited font-sizes for both text fields and buttons**

image

gustavkj (commented):
> One thing I am unsure about is the additional confirmation dialog if there are 10+ links. It will quickly annoy regular users with one unnecessary extra click.

True, but it is also possible to interact via the keyboard (i.e. `Enter` to confirm or `Escape` to cancel). I think that is usually how I interact with those kinds of confirmation boxes. But that might not be how the majority will interact with it. 🤔 

I like having it as, similar to you, I feel like the browser kind of get hijacked when opening many tabs. So, having gotten a clear warning allowing you to cancel the action feel like a kind gesture to the user.

But I can remove it also. Wdyt? If it it turns out to annoy users, we there'll probably be issues created about it. (Same goes for if the confirmation box is annoying).

> And a huge thank you for adding a few test cases for the MB edit link generation, these are really appreciated.

I'll keep it up 👍

server/components/LinkWithMusicBrainz.tsx
kellnerd: > Personally, I try to avoid non-null assertions most often as they can hide null or undefined values that should be checked for.
> 
> In this case, I would prefer to refactor the actions page to make it clearer what data is expected.

I fully agree with that. If TS was a bit smarter and would understand that a dependent pair of variables (`release` and `releaseUrl`) are either both defined or undefined, the assertions wouldn't be necessary.

But I prefer the TS assertion marker over a redundant extra condition which is checked in JS every time. So I've only used the assertion marker in those cases where I already had a condition to check whether `release` is truthy.

So what are you suggesting as an improvement? Replacing the additional variable `releaseUrl` (and maybe also `releaseMbid`) by a new property of `release`?

server/routes/release/actions.tsx
```diff
@@ -197,14 +197,14 @@ export default defineRoute(async (req, ctx) => {
 				<LinkWithMusicBrainz
 					entities={allArtists}
 					entityType='artist'
-					sourceEntityUrl={releaseUrl}
+					sourceEntityUrl={releaseUrl!}
```
kellnerd: This assertion is not safe since we have no condition around this component at all, we need a condition on `releaseUrl` here.

P.S. I just saw that I previously had such a condition, although the `allArtists.map(...)` wrapper around the component is technically sufficient. (When there is no release, we have an empty artist array.)
The same construct for `allRecordings` did not have a condition but an assertion, so this is probably the inconsistency that you meant...

server/routes/release/actions.tsx
```diff
@@ -224,7 +224,7 @@ export default defineRoute(async (req, ctx) => {
 				<LinkWithMusicBrainz
 					entities={allRecordings}
 					entityType='recording'
-					sourceEntityUrl={releaseUrl}
+					sourceEntityUrl={releaseUrl!}
```
kellnerd: This assertion is not safe since we have no condition around this component at all, we need a condition on `releaseUrl` here.

server/static/harmony.css
```diff
@@ -202,13 +204,14 @@ input[type='submit'], button {
...
 	border-radius: 0.3em;
+	font-size: inherit;
+	font-family: inherit;
 }
 input[type='submit']:active, button:active {
 	border-style: inset;
 }
 input[type='submit']:focus, button:focus, input[type='submit']:hover, button:hover {
 	background-color: var(--button-focus);
-	cursor: pointer;
```
kellnerd, line 208: This cursor style was specifically [requested](https://github.com/kellnerd/harmony/issues/51#issuecomment-3200831938), so we shouldn't change it.
I only suggested to remove it from the new `button.open-all-links` rules since those already inherit the general rules for buttons.

server/static/harmony.css
kellnerd: Hmm, I'm not sure about the text inputs, I kinda prefer the current smaller font for those (`font-size: smaller` if we want to make it more explicit).
But I don't want to further delay this PR with general design questions, I was mainly concerned by the redundant CSS properties that you have since removed.
Let's just revert the font-size change for general inputs and buttons, I agree to have the new open all buttons styled differently for now and am happy to discuss further CSS changes separately.

server/static/harmony.css
arsinclair: Thank you for keeping it afloat. Yes, a pointer cursor should be used for links and buttons, this is a UI convention. 

For other interactive elements, such as checkbox, select, radiobutton, etc. this is not a convention, but personally I find it more intuitive if the cursor also changes to a pointer.

server/components/LinkWithMusicBrainz.tsx
gustavkj: > So what are you suggesting as an improvement? Replacing the additional variable `releaseUrl` (and maybe also `releaseMbid`) by a new property of `release`?

Kind of, initial thought was that there could be a function preparing the content. That way, the return type of the function could be a union of the different expected types (one with `release`, `releaseMbid`, and `releaseUrl` set, and one with all optional or `undefined`). 🤔 This could also open up for adding testing of the data preparation for the actions page without having to deal with the visual components.

But that's something for another PR, if it sounds reasonable. (I could make a proof of concept if it sounds interesting).

server/static/harmony.css
gustavkj: I've changed it back now. 🙂

gustavkj:
> Unless he wants to spend further time on it, I think this is ready for a first release.

I'm sure we can find improvements or ways to make it work better with the user scripts. But I feel like as an initial implementation, it is good enough. 🙂

kellnerd (approved):
Thank you for contributing the most wanted new feature, according to [GitHub upvotes](https://github.com/kellnerd/harmony/issues?q=is%3Aissue%20sort%3Areactions-%2B1-desc) :open_mouth:
And sorry that this took surprisingly long for a task which I labeled as good first issue.

server/components/LinkWithMusicBrainz.tsx
kellnerd: That sounds sensible to me, probably better than fighting with a new type that extends `HarmonyRelease`.

Issue #125, opened 2025-07-06
Hide ISRC submission links if ISRCs were already added


kellnerd:
This is currently blocked because the cached MB API responses don't contain ISRCs to compare against. Bad design decision in retrospect, but this can be addressed once I've released the refactoring which supports checking multiple legacy API URLs in the cache.

So this essentially depends on https://github.com/kellnerd/harmony/issues/121.

PR #151, opened 2025-10-07
feat: Hide ISRC submission links if ISRCs were already added

gustavkj:
Closes: https://github.com/kellnerd/harmony/issues/125

Hides the ISRC submission section if the recordings already have the ISRCs set.

Some examples to test:

- Already have the ISRCs:
	- `/release/actions?release_mbid=de25aec6-9340-44e5-8584-b3571ca4a7ac`
	- `/release/actions?release_mbid=6698da2f-8c05-4765-8bf4-4fffe0cebaec`
- Missing ISRCs:
	- `/release/actions?musicbrainz=b05b6d2c-8667-470b-b8c0-9581fd2027cb&spotify=70vKZBocKuuWTT08nhVLj8&gtin=886449193363&itunes=&tidal=&region=SE&release_mbid=b05b6d2c-8667-470b-b8c0-9581fd2027cb`
	- `/release/actions?release_mbid=24452979-7205-448d-811f-d03cb649ae3c`
- Missing a lot of ISRCs: `/release/actions?spotify=2Nqpb6pY4bpsfdTfBavVng&gtin=00602445196982&deezer=&itunes=&region=SE&release_mbid=157101a5-f0b7-4d98-b9b2-5ad38a5a4820`

### DIFF
--- a/server/components/ISRCSubmission.tsx
+++ b/server/components/ISRCSubmission.tsx
@@ -1,15 +1,68 @@
+import { SpriteIcon } from './SpriteIcon.tsx';
+
 import { musicbrainzTargetServer } from '@/config.ts';
-import type { HarmonyRelease } from '@/harmonizer/types.ts';
+import type { HarmonyRelease, HarmonyTrack, ProviderInfo } from '@/harmonizer/types.ts';
 import { join } from 'std/url/join.ts';
+import type { EntityWithMbid } from '@kellnerd/musicbrainz/api-types';
 
-export function MagicISRC({ release, targetMbid }: { release: HarmonyRelease; targetMbid?: string }) {
-	const allTracks = release.media.flatMap((medium) => medium.tracklist);
+// TODO: incomplete type, expose a suitable type from @kellnerd/musicbrainz?
+export interface EntityWithISRCs extends EntityWithMbid {
+	isrcs?: string[];
+}
+
+function getISRCProvider(release: HarmonyRelease): ProviderInfo | undefined {
 	const isrcSource = release.info.sourceMap?.isrc;
-	const isrcProvider = release.info.providers.find((provider) => provider.name === isrcSource);
-	if (!(allTracks.some((track) => track.isrc) && isrcProvider)) {
-		return null;
-	}
+	if (!isrcSource) return undefined;
+	return release.info.providers.find((provider) => provider.name === isrcSource);
+}
 
+export function ISRCSubmission(
+	{ release, targetMbid, recordingsCache }: {
+		release: HarmonyRelease;
+		targetMbid?: string;
+		recordingsCache?: EntityWithISRCs[];
+	},
+) {
+	const isrcProvider = getISRCProvider(release);
+
+	if (!isrcProvider) return null;
+	const allTracks = release.media.flatMap((medium) => medium.tracklist);
+
+	const needsSubmission = (track: HarmonyTrack): boolean => {
+		// If there's no ISRC, nothing to submit
+		if (!track.isrc) return false;
+
+		const recordingMbid = track.recording?.mbid;
+		// If there's no recording MBID, consider the ISRC as new (needs submission)
+		if (!recordingMbid) return true;
+
+		const mbEntity = recordingsCache?.find((e) => e.id === recordingMbid);
+		// If MB entity is missing or doesn't include the ISRC, it's new
+		return !mbEntity?.isrcs?.includes(track.isrc);
+	};
+
+	const noPendingISRCSubmissions = !allTracks.some(needsSubmission);
+
+	if (noPendingISRCSubmissions) return null;
+
+	return (
+		<div class='action'>
+			<SpriteIcon name='disc' />
+			<p>
+				<MagicISRC allTracks={allTracks} targetMbid={targetMbid} isrcProvider={isrcProvider} />
+				: Submit ISRCs from <a href={isrcProvider.url}>{isrcProvider.name}</a> to MusicBrainz
+			</p>
+		</div>
+	);
+}
+
+function MagicISRC(
+	{ allTracks, targetMbid, isrcProvider }: {
+		allTracks: HarmonyTrack[];
+		targetMbid?: string;
+		isrcProvider: ProviderInfo;
+	},
+) {
 	const query = new URLSearchParams(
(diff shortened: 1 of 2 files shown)


server/routes/release/actions.tsx
```diff
@@ -186,13 +191,12 @@ export default defineRoute(async (req, ctx) => {
...
-							<MagicISRC release={release} targetMbid={releaseMbid!} />
-							: Submit ISRCs from <a href={isrcProvider.url}>{isrcProvider.name}</a> to MusicBrainz
-						</p>
-					</div>
+					<ISRCSubmission
+						release={release}
+						targetMbid={releaseMbid}
+						isrcProvider={isrcProvider}
+						recordingsCache={mbRecordings}
+					/>
```
kellnerd, line 192: With your refactoring `isrcProvider` is no longer needed outside this component and doesn't have to be passed in. We can infer it from `release` inside the component now.
But over there we can pass it along to the now internal component `MagicISRC` and avoid computing it twice.

server/components/ISRCSubmission.tsx
```diff
@@ -1,8 +1,49 @@
...
+		targetMbid?: string;
+		isrcProvider?: ProviderInfo;
+		recordingsCache?: EntityWithISRCs[];
+	},
+) {
+	if (!isrcProvider || !release) return null;
+	const allTracks = release.media.flatMap((medium) => medium.tracklist);
+	if (!allTracks.some((track) => track.isrc)) return null;
+
+	const noNewISRCs = allTracks.every((track) => {
```
kellnerd: Something seems to be wrong with this logic, the negation of the variable name makes it very confusing to follow what has to be returned from the `every` callback...
I think it's currently wrong and would have to use `allTracks.some` if there is no outer negation, according to De Morgan's laws.

But I would suggest to rewrite this to have the callback check whether the track has a recording MBID and a new ISRC.
Then we can simply negate the result of `every`, which is much simpler to follow mentally.

server/components/ISRCSubmission.tsx
```diff
@@ -1,8 +1,49 @@
 import { musicbrainzTargetServer } from '@/config.ts';
-import type { HarmonyRelease } from '@/harmonizer/types.ts';
+import type { HarmonyRelease, ProviderInfo } from '@/harmonizer/types.ts';
 import { join } from 'std/url/join.ts';
+import { SpriteIcon } from './SpriteIcon.tsx';
```
kellnerd: Please give the TSX import a separate section at the top of the module.

server/routes/release/actions.tsx
gustavkj: Good point 👍

server/components/ISRCSubmission.tsx
gustavkj: For every track, it checks if its ISRC exist in the array of ISRCs for the associated recording. If the ISRC of every track is already present, it is true. If any ISRC is missing (or any of the look ups fail), it returns false.

So, if it is true, there are no new ISRCs to add and the action is not shown. If it is false, there are new ISRCs to add (or the look up failed and there might be new ISRCs to add), and therefore it is shown.

The name `noNewISRCs`, could also be `isrcsAreUpToDate` or `noPendingISRCSubmissions`. 🤔 

I'll change it to use `some` instead. I think the outcome/return is the same, but `some` is more efficient since it doesn't look through all elements. 👍

server/components/ISRCSubmission.tsx
```diff
@@ -36,23 +44,22 @@ export function ISRCSubmission(
...
+function MagicISRC(
+	{ allTracks, targetMbid, isrcProvider }: {
+		allTracks: HarmonyTrack[];
+		targetMbid?: string;
+		isrcProvider: ProviderInfo;
+	},
+) {
 	const query = new URLSearchParams(
-		allTracks.map((track, index) => [`isrc${index + 1}`, track.isrc ?? '']),
+		allTracks.map((track, index) => [`isrc${index + 1}`, track.isrc!]),
```
kellnerd: We can't do that since we have only checked whether some track has an ISRC. It is totally possible that some tracks may not have an ISRC.

server/components/ISRCSubmission.tsx
kellnerd: Thank you, the new comments and the named callback make it much easier to follow and the check should be faster as well.
I wrote down the following notes to compare the old and the new logic:

```
I = track has an unknown ISRC (or no cached recording or no cached ISRC)
not I = track has known ISRC (and a cached recording and a cached ISRC)
M = track's recording has an MBID, we can check for known ISRCs
not M = recording has no MBID, we can't check for known ISRCs
S = pending ISRC submissions (potentially)

Intuitively (for me): S = some(I or not M)
Currently: not S = not some(not M or I) = every(not(not M or I)) = every(M and not I)
Previously: not S = every(M and not I)
```

So your previous code was indeed correct and equivalent to the current code, my mistake was that I didn't get the consequence of a missing MBID yesterday.
Somehow I thought "no MBID, nothing to submit" but we don't need the recording MBID to submit ISRCs for the release... And in reality we will always have a recording MBID even, since the combined release on the actions page has always been merged with data from the MB provider, which provides recording MBIDs.

server/components/ISRCSubmission.tsx
```diff
@@ -29,16 +29,19 @@ export function ISRCSubmission(
...
-
-		if (!recordingMbid) return false;
+		// If there's no recording MBID, consider the ISRC as new (needs submission)
+		if (!recordingMbid) return true;
 
 		const mbEntity = recordingsCache?.find((e) => e.id === recordingMbid);
-		return mbEntity?.isrcs?.includes(track.isrc!);
-	});
+		// If MB entity is missing or doesn't include the ISRC, it's new
+		return !mbEntity?.isrcs?.includes(track.isrc!);
```
kellnerd: Instead of asserting that the ISRC is defined here, we could have an early exit condition for it.
If there is no ISRC, we don't have anything to submit for this track.

gustavkj: Yeah, it's more efficient to integrate it into the `needsSubmission` function than having the separate `some` (and also re-add the nullish coalescing operator in `MagicISRC`).

server/components/ISRCSubmission.tsx
gustavkj: Right, I thought it was the other way around 🤦

Issue #112, opened 2025-06-01
Full size cover art images


Ser-Erate:
The release actions page has links to the cover art of releases from various sites, however, while it typically has the maximum resolution from Spotify and I believe Apple Music, it does not have them from Tidal or Deezer, instead having the default 1200x1200 and 1000x1000, Respectively. However, these are not the maximum sizes on these sites; I've seen Deezer images as big as 1900x1900 (They may go higher, but I can't personally attest to it), and Tidal can apparently be up to 5000x5000. What's more, in both cases, the maximum size can be easily accessed via URL manipulation, so it shouldn't be hard to implement links to the maximum size in Harmony.
On Deezer, the address for the cover art typically ends with `/1000x1000-000000-80-0-0.jpg` (For a 1000x1000 image), but you can access the maximum size by replacing "1000x1000" with "1900x1900" (I'm not certain this is the largest size, though from my experience, 1900x1900 will give you 1900x1900 if available, And the biggest size if it's smaller, Whereas 2000x2000 typically shows no image and says "Unauthorized")
On Tidal, the address for cover art typically ends with `1280x1280.jpg` (For 1280x1280 images), but you can access the maximum size by replacing "1280x1280" with "origin"
Examples: [1000x1000 Deezer image linked from Harmony](https://cdn-images.dzcdn.net/images/cover/4b58680a78a4669e9bcb145112d2dbf9/1000x1000-000000-80-0-0.jpg) vs [1900x1900 acquired by changing the URL.](https://cdn-images.dzcdn.net/images/cover/4b58680a78a4669e9bcb145112d2dbf9/1900x1900-000000-80-0-0.jpg)
[1280x1280 Tidal image linked from Harmony](https://resources.tidal.com/images/5f7a62e9/f561/46ba/a7ed/63df6225e787/1280x1280.jpg) vs [5000x5000 acquired by changing the URL.](https://resources.tidal.com/images/5f7a62e9/f561/46ba/a7ed/63df6225e787/origin.jpg)

Ser-Erate:
This could be especially useful in cases where Apple Music has abnormally small maximum size, For example on [this release](https://musicbrainz.org/release/37d9d860-7525-4a2d-afad-9f8b73bf7d4d), The maximum size on Apple Music is 620x620, but Deezer goes up to 1200x1200, And Tidal has 1400x1400.

kellnerd:
[...]

PR #156, opened 2025-10-13
feat: Soundcloud Support

### LINKED ISSUE #35: SoundCloud
I know it's mentioned in #5, but I figured I'd start up a ticket with a link to the API docs at least~

https://developers.soundcloud.com/docs

a couple notes about SoundCloud:
- some title cleanup might be needed? especially stuff like "[FREE DL IN DESCRIPTION]", which is quite common
- GTIN and ISRC are optional metadata
- playlists and albums are both implemented as "sets", and perhaps both should be importable? I know a lot of artists don't properly set the type for what's pretty clearly an album (even [Taylor Swift](https://soundcloud.com/taylorswiftofficial/sets) has some such examples)
- Creative Commons Licenses should be detected and added when present
- track downloads are optional and should be detected (sometimes with a Buy link that goes to a file hosting service like MediaFire or MEGA, or with a DL LINK IN DESCRIPTION)

Lioncat6:
Closes https://github.com/kellnerd/harmony/issues/35

SoundCloud currently implements the following:

- Looking up releases as a set/playlist _or_ track
  - Unlike Bandcamp, there is no way to determine if a track has a parent playlist/album with the v1 API
- Getting track ISRC, duration, author, and region information (Though, I haven't found a track, yet that has a set region restriction)
- Track/Playlist publish date, falling back to the creation date if not present
- Track and album artwork
- Getting playlist type
- Determining playlist label from tracks


providers/SoundCloud/mod.ts
```diff
@@ -0,0 +1,470 @@
...
+
+	readonly trackUrlPattern = new URLPattern({
+		hostname: 'soundcloud.com',
+		pathname: '/:artist/:title',
+	});
+
+	readonly artistUrlPattern = new URLPattern({
+		hostname: 'soundcloud.com',
+		pathname: '/:artist',
+	});
```
kellnerd, line 46: Once we have test cases for URL extraction, we can try to combine these URL patterns.

providers/SoundCloud/mod.ts
```diff
@@ -0,0 +1,470 @@
...
+		}
+	}
+
+	override getLinkTypesForEntity(entity: EntityId): LinkType[] {
+		if (entity.type === 'track') {
+			// All tracks offer free streaming, but some don't have free downloads enabled.
+			return ['free streaming'];
+		}
+		// MB has special handling for Bandcamp artist URLs
+		return ['discography page'];
```
kellnerd: I have to check which link types MB uses for SC, but this looks like a leftover from BC? ([special URL handling is here](https://github.com/kellnerd/harmony/blob/7289e99599c8e86594285c0477ac44dec8a9f1e6/musicbrainz/seeding.ts#L143))

providers/SoundCloud/mod.ts
```diff
@@ -0,0 +1,470 @@
...
+		super('SoundCloud', details?.status, url); //While there exists a message field in the error response, it's usually empty, despite status being deprecated.
+	}
+}
+
+export class SoundCloudReleaseLookup extends ReleaseLookup<SoundCloudProvider, RawReponse> {
+	rawReleaseUrl: URL | undefined;
+
+	constructReleaseApiUrl(): URL | undefined {
+		return undefined;
+	}
```
kellnerd: This method should contain parts of the logic which are currently included in `getRawRelease`.
It is also used to construct the API URL which is shown in the provider section of the release lookup page.

providers/SoundCloud/mod.ts
```diff
@@ -0,0 +1,470 @@
...
+			const release: HarmonyRelease = {
+				title: playlistResponse.title,
+				artists: [this.makeArtistCredit(playlistResponse.user)],
+				externalLinks: [{
+					url: rawRelease.href,
+					types: this.getReleaseTypes(playlistResponse),
+				}],
+				media: [{
+					format: 'Digital Media',
+					tracklist: playlistResponse.tracks?.filter(isNotNull).map((track, idx) => this.convertRawTrack(track, idx)) || [],
```
kellnerd:
```suggestion
					tracklist: playlistResponse.tracks?.filter(isNotNull).map(this.convertRawTrack.bind(this)) ?? [],
```
probably

providers/SoundCloud/mod.ts
```diff
@@ -0,0 +1,470 @@
...
+				for (const track of playlist.tracks) {
+					if (!track.downloadable) {
+						allDownloadable = false;
+						break;
+					}
+				}
+				if (allDownloadable) {
+					types.push('free download');
+				}
+			}
```
kellnerd:
```suggestion
			if (playlist.tracks?.every((track) => track.downloadable)) {
				types.push('free download');
			}
```

providers/SoundCloud/mod.ts
```diff
@@ -0,0 +1,470 @@
...
+		return [];
+	}
+
+
+	getArtwork(release: SoundcloudPlaylist | SoundcloudTrack): Artwork[] | undefined {
+		const artworks: Artwork[] = [];
+		const artworkUrl = release.artwork_url?.replace(/-(large|medium|small)\./, '-t300x300.');;
+		if (artworkUrl) {
+			// Replace large/medium/small with t500x500 to get the largest available image.
+			const artworkFull = artworkUrl.replace(/-(large|medium|small)\./, '-t500x500.');
```
kellnerd: These two replacements are conflicting, the second will never do anything. I would inline the two replacements into the `artworks.push` call.
Don't worry too much about image sizes, that's #112. Everything is fine as long as the thumbnails aren't too large.

providers/SoundCloud/mod.ts
```diff
@@ -0,0 +1,470 @@
...
+			if (playlist.label_name) {
+				const label:Label = {
+					name: playlist.label_name,
+				}
+				return [label];
+			} else {
+				if (playlist.tracks?.length > 0) {
+					const labelName = playlist.tracks[0].label_name;
+					for (const track of playlist.tracks) {
+						if (track.label_name !== track.label_name) {
```
kellnerd:
```suggestion
						if (track.label_name !== labelName) {
```
is probably what you intended to do.

providers/SoundCloud/mod.ts
```diff
@@ -0,0 +1,470 @@
...
+							// Different label names found, cannot determine a single label for the release.
+							return undefined;
+						}
+					}
+					if (labelName) {
+						const label:Label = {
+							name: labelName,
+						}
+						return [label];
+					}
```
kellnerd: Alternative implementation (`return undefined` is implicit):
```suggestion
					if (labelName && playlist.tracks.every((track) => track.label_name === labelName)) {
						return [{
							name: labelName,
						}];
					}
```

providers/SoundCloud/mod.ts
```diff
@@ -0,0 +1,470 @@
...
+					name: track.label_name,
+				}
+				return [label];
+			}
+		}
+		return undefined;
+	}
+
+	getReleaseDate(release: SoundcloudTrack | SoundcloudPlaylist): PartialDate | undefined {
+		if (release.release_day, release.release_month, release.release_year) {
```
kellnerd: This would only check whether the year is truthy and ignore month and day.
```suggestion
		if (release.release_year || release.release_month || release.release_day) {
```

providers/SoundCloud/mod.ts
```diff
@@ -0,0 +1,470 @@
...
+				return parseISODateTime(date.toISOString());
+			}
+		}
+		return undefined;
+	}
+
+	parseSoundcloudTimestamp(timestamp: string): Date | undefined {
+		const segments = timestamp.split(' ');
+		if (segments.length === 3) {
+			// Convert "2025/07/30 07:13:31 +0000" to "2025-07-30T07:13:31+00:00"
```
kellnerd: I have to look at some examples, but if these are always in UTC we could simply extract day, month and year and don't have to worry about the rest.

providers/SoundCloud/mod.ts
```diff
@@ -0,0 +1,470 @@
...
+	}
+
+	override readonly launchDate: PartialDate = {
+		year: 2008,
+		month: 10,
+		day: 17,
+	};
+
+	override readonly features: FeatureQualityMap = {
+		'cover size': 500,
```
kellnerd, line 123: I was able to obtain bigger images (about 1500px) with Image Max URL, so we should check that again.

kellnerd (changes requested):
Thank you for looking into SoundCloud!
Although this is still a draft and I haven't made an account to obtain API credentials yet, I couldn't resist to review the code once.
I have left a few comments but I guess there will be more once I get to play with the v1 (and maybe v2) API.

P.S. CI is complaining about code formatting, see [here](https://github.com/kellnerd/harmony/blob/main/CONTRIBUTING.md#code-submission) how to reproduce that locally.

providers/SoundCloud/mod.ts
Lioncat6: 500x500 is the highest we can get for sure. Using `original` in the url instead of `t500x500`, we can get the original file, which I've personally seen up to 3000x3000, the ending of the url changes depending on what the artist uploaded the file as: either .jpg or .png

Lioncat6: Update, just found one at 3071x3071.

Lioncat6: Further update: After messing around with my own images, there is no image resolution limit, but compression is applied to whatever you upload

providers/SoundCloud/mod.ts
Lioncat6:
```
if (url?.hostname.replace('www.','') == 'soundcloud.com') {
	return typeIds.soundcloud;
}
```
Added special handling for soundcloud urls

providers/SoundCloud/mod.ts
```diff
@@ -0,0 +1,438 @@
...
+import { ProviderError, ResponseError } from '@/utils/errors.ts';
+import { getFromEnv } from '@/utils/config.ts';
+import { isNotNull } from '@/utils/predicate.ts';
+import { ApiError, RawReponse, SoundcloudPlaylist, SoundcloudTrack, SoundcloudUser } from './api_types.ts';
+import { encodeBase64 } from 'std/encoding/base64.ts';
+import { ResponseError as SnapResponseError } from 'snap-storage';
+import { capitalizeReleaseType } from '../../harmonizer/release_types.ts';
+
+const soundcloudClientId = getFromEnv('HARMONY_SOUNDCLOUD_CLIENT_ID') || '';
+const soundcloudClientSecret = getFromEnv('HARMONY_SOUNDCLOUD_CLIENT_SECRET') || '';
```
gustavkj, line 29: Might be nice to add a link to e.g. `https://developers.soundcloud.com/docs/api/guide` in a comment above these (similar to in the Spotify and Tidal providers). 🙂 Also, it would be good to add these env vars to the `.env.example` file, so it is easy to know which env vars are needed for local development.

providers/SoundCloud/mod.ts
Lioncat6: Added tests, haven't cached the request data for them because I don't know how to do it for a specific URL.

providers/SoundCloud/mod.ts
Lioncat6: Done! https://github.com/kellnerd/harmony/pull/156/commits/df02363a06237fedba83d74041c7b9e25fd0e117

providers/SoundCloud/mod.ts
Lioncat6: Since there is no way to determine the url extension for the `-original` urls, I think it should stay with the `t500x500` URL and let the image importer userscript handle it.

Otherwise, a request could be made here to determine the extension
https://github.com/kellnerd/harmony/blob/bf0554bab20c2f78f8db5a3a0b2f5d80b17e7677/providers/SoundCloud/mod.ts#L342-L355

musicbrainz/seeding.ts
```diff
@@ -145,6 +145,9 @@ export function convertLinkType(entityType: EntityType, linkType: LinkType, url?
 			if (url?.hostname.endsWith('.bandcamp.com')) {
 				return typeIds.bandcamp;
 			}
+			if (url?.hostname.replace('www.', '') == 'soundcloud.com') {
```
kellnerd, line 148: Do the URLs ever have a www subdomain? The provider implementation doesn't accept it at least.

providers/SoundCloud/__snapshots__/mod.test.ts.snap
```diff
@@ -0,0 +1,102 @@
...
+            ],
+          },
+          title: "Piercing Light (Mako Remix)",
+        },
+      ],
+    },
+  ],
+  packaging: "None",
+  releaseDate: {
+    day: 12,
```
kellnerd, line 93: The website says 14 January 2016 when I hover the "9 years ago". Even if there is a timezone offset, it can't be two days?
We should investigate all the available date fields, ideally for a release with a known release date (for example from another platform).

providers/SoundCloud/api_types.ts
```diff
@@ -1,98 +1,95 @@
...
-	| 'mini';
+export type SoundcloudImageFormats = 't500x500';
+'crop';
+'t300x300';
+'large';
+'t67x67';
+'badge';
+'small';
+'tiny';
+'mini';
```
kellnerd: This union type and a few others are broken now, please add the pipe separators back.

providers/SoundCloud/mod.ts
kellnerd: I'm sorry if I've been unclear, but I wanted to express that it is simpler (and less error-prone) to just extract a `PartialDate` directly from the string instead of creating a `Date` instance, serializing it and parsing it again.

kellnerd (changes requested):
Thank you for adding a few provider tests! For today I have a couple of questions and a few refactorings which went wrong.

providers/SoundCloud/mod.ts
Lioncat6: Simplified date logic

providers/SoundCloud/__snapshots__/mod.test.ts.snap
Lioncat6: As it turns out, SoundCloud entities have 4 "release" dates... the most reliable being the dedicated release fields, as those are manually entered by the artist (usually pretty rarely, similar to ISRC and UPC data). The date that appears on the website is the `display_date` field, which mirrors the `last_modified` date. The date I'm currently using as a fallback for the actual release date is the `created_at` date, which isn't what's displayed on the website.

providers/SoundCloud/mod.ts
```diff
@@ -393,23 +393,22 @@ export class SoundCloudReleaseLookup extends ReleaseLookup<SoundCloudProvider, R
...
-			const date = new Date(iso);
-			if (!isNaN(date.getTime())) {
-				return date;
-			}
+			// Timestamp format "2025/07/30 07:13:31 +0000"
+			const dateSegments = segments[0].split('/');
+			const date: PartialDate = {
+				year: Number(dateSegments[0]) || undefined,
+				month: Number(dateSegments[1]) || undefined,
+				day: Number(dateSegments[2]) || undefined,
```
kellnerd:
```suggestion
				year: Number.parseInt(dateSegments[0]),
				month: Number.parseInt(dateSegments[1]),
				day: Number.parseInt(dateSegments[2]),
```

kellnerd:
I have taken the liberty to finally push my changes to your branch.
The commit messages should speak for themselves, but please ask if you have any question or doubts.

Sorry that it took me so long, these changes are already 2-3 weeks old, lol.
I though that it would be able to quickly fix the following issues as well, but I didn't:
- [ ] **Link types** are sometimes incorrect
   - Not all tracks of all releases are streamable for example
   - There is a property `streamable` at playlist and track level which probably helps
   - I would rather ship an initial version which doesn't set link types than a version which results in inaccurate data
- [ ] **Release type** handling has to be safeguarded (PLAYLIST is not a valid type on MB for example)
- [ ] **Cover size**
   - [ ] Feature value doesn't matter too much (it is only used to rank providers), but the average should be higher than 500 (1500? 3000?)
   - [ ] I would still return "-original.*" URLs (unless "-original.jpg" is invalid when the original is PNG)
- [ ] Most importantly: The API returns **empty track lists** for all major artists that I've tried!
   - This is unrelated to the Go+ premium subscription
   - Not even the specific `/playlists/$ID/tracks` endpoint helps

Serial-ATA:
@Lioncat6 Need any help moving this forward? I've been adding a lot of soundcloud releases lately, would be great to automate it

Lioncat6:
@Serial-ATA I need to double check if my repo is up to date with kelnerd's changes in the soundcloud branch of this one

mjsir911:
I've ended up implementing the same in [dcf3cdc](https://github.com/mjsir911/harmony/commit/dcf3cdc28682d580325332fc88e82b88fbc811fb) but this all seems more fleshed out

PR #168, opened 2025-11-21
feat(mora): add provider for mora.jp

### LINKED ISSUE #126: Add support for mora.jp
mora.jp often has barcodes in url like here
https://mora.jp/package/43000006/00602567550631/

and sometimes catalog number like here
https://mora.jp/package/43000188/G010005563453V/

would be great to add this swebsite to harmony

Serial-ATA:
closes #126

Draft for now, since it could probably use more tests

### DIFF
--- a/providers/Mora/json_types.ts
+++ b/providers/Mora/json_types.ts
@@ -21,10 +21,10 @@ export interface PackageMeta {
 }
 
 export interface Track {
-	arranger?: string;
-	composer?: string;
+	arranger: string | null;
+	composer: string | null;
 	/** Lyricist name */
-	lyrics?: string;
+	lyrics: string | null;
 
 	artistName: string;
 	artistNo: number;


providers/Mora/json_types.ts
```diff
@@ -0,0 +1,54 @@
...
+	fullsizeimage: string;
+	title: string;
+	labelCode: string;
+	labelcompanyname: string;
+	master: string;
+	distPartNo: string;
+	startDate: string;
+	mediaFormatNo: MediaFormat;
+	mediaType: MediaType;
+	trackList: Record<number, Track>;
```
kellnerd:
```suggestion
	trackList: Track[];
```

providers/Mora/mod.ts
```diff
@@ -0,0 +1,251 @@
+import { ApiArgs, MediaFormat, PackageMeta, Track, WithApiUrl } from './json_types.ts';
```
kellnerd:
```suggestion
import type { ApiArgs, MediaFormat, PackageMeta, Track, WithApiUrl } from './json_types.ts';
```
avoids module reloading for type-only changes.

providers/Mora/mod.ts
```diff
@@ -0,0 +1,251 @@
...
+	HarmonyTrack,
+	LinkType,
+	ReleaseGroupType,
+} from '@/harmonizer/types.ts';
+import { type CacheEntry, MetadataProvider, ReleaseLookup } from '@/providers/base.ts';
+import { DurationPrecision, FeatureQuality, FeatureQualityMap } from '@/providers/features.ts';
+import { parseISODateTime, PartialDate } from '@/utils/date.ts';
+import { ProviderError, ResponseError } from '@/utils/errors.ts';
+import { extractMetadataTag } from '@/utils/html.ts';
+import { isValidGTIN } from '../../utils/gtin.ts';
```
kellnerd:
```suggestion
import { isValidGTIN } from '@/utils/gtin.ts';
```

providers/Mora/mod.ts
```diff
@@ -0,0 +1,251 @@
...
+		'cover size': 200,
+		'duration precision': DurationPrecision.SECONDS,
+		'GTIN lookup': FeatureQuality.MISSING,
+		'MBID resolving': FeatureQuality.PRESENT,
+		'release label': FeatureQuality.GOOD,
+	};
+
+	readonly entityTypeMap = {
+		artist: 'artist',
+		release: 'album',
```
kellnerd: Maybe we should also the release type "package"? They seem to use "singles/albums" in one place, so they are not really misusing album to mean of any kind of release (as many other providers do).

providers/Mora/mod.ts
```diff
@@ -0,0 +1,251 @@
...
+	readonly releaseLookup = MoraReleaseLookup;
+
+	constructUrl(entity: EntityId): URL {
+		if (entity.type === 'artist') {
+			return new URL(`https://mora.jp/artist/${entity.id}/`);
+		} else if (entity.type === 'album') {
+			return new URL(`https://mora.jp/package/${entity.id}/`);
+		}
+
+		throw new ProviderError(this.name, `Incomplete release ID '${entity.id}' does not match format \`band/title\``);
```
kellnerd: Probably a leftover from Bandcamp :grin:
When we use mora's names "artist" and "package" for the types, we can simplify this method and not worry about invalid types here. (Many provider implementations currently assume that `EntityId` objects are valid.)

providers/Mora/mod.ts
```diff
@@ -0,0 +1,251 @@
...
+			};
+		}
+
+		return undefined;
+	}
+
+	override parseProviderId(id: string, entityType: HarmonyEntityType): EntityId {
+		return { id, type: this.entityTypeMap[entityType] };
+	}
+
```
kellnerd: Since there is no custom ID serialization (to support multiple release types with their own IDs) we also don't need to override the de-serialization here.
```suggestion
```

providers/Mora/mod.ts
```diff
@@ -0,0 +1,251 @@
...
+		return this.fetchJSON<ApiArgs>(webUrl, {
+			policy: { maxTimestamp },
+			responseMutator: async (response) => {
+				const html = await response.text();
+				const metaTag = extractMetadataTag(html, 'msApplication-Arguments');
+				if (!metaTag) {
+					throw new ResponseError(this.name, 'Response is missing the expected <meta> tag', webUrl);
+				}
+
+				const apiArgsRaw = metaTag.replace(/&quot;/g, '"');
```
kellnerd: You shouldn't need this, `extractMetadataTag` already decodes HTML entities?

providers/Mora/mod.ts
```diff
@@ -0,0 +1,251 @@
...
+
+		return release.data;
+	}
+
+	convertRawRelease(albumPage: PackageMeta): HarmonyRelease {
+		const label = { name: albumPage.labelcompanyname };
+		const tracklist = Object.entries(albumPage.trackList).map(([index, track]) =>
+			this.convertRawTrack(Number(index), track)
+		);
+		const types: ReleaseGroupType[] = [Object.keys(albumPage.trackList).length > 1 ? 'Album' : 'Single'];
```
kellnerd: We shouldn't make assumptions here, other providers only fill this if the API provides it. (Whether the APIs are accurate is another question.)

providers/Mora/mod.ts
```diff
@@ -0,0 +1,251 @@
...
+			copyright: albumPage.master,
+			info: this.generateReleaseInfo(),
+		};
+
+		return release;
+	}
+
+	convertRawTrack(index: number, rawTrack: Track): HarmonyTrack {
+		return {
+			number: index,
```
kellnerd: Tracks have a `trackNo `property which we should use, we shouldn't need the zero-based index.
```suggestion
	convertRawTrack(rawTrack: Track): HarmonyTrack {
		return {
			number: rawTrack.trackNo,
```

providers/Mora/mod.ts
```diff
@@ -0,0 +1,251 @@
...
+		this.updateCacheTime(timestamp);
+
+		return release.data;
+	}
+
+	convertRawRelease(albumPage: PackageMeta): HarmonyRelease {
+		const label = { name: albumPage.labelcompanyname };
+		const tracklist = Object.entries(albumPage.trackList).map(([index, track]) =>
+			this.convertRawTrack(Number(index), track)
+		);
```
kellnerd:
```suggestion
		const tracklist = albumPage.trackList.map((track, index) =>
			this.convertRawTrack(index, track)
		);
```
would be the way to pass on the index, but since we don't need it, we can drop it from the methods parameters and simplify this further to
```suggestion
		const tracklist = albumPage.trackList.map(this.convertRawTrack.bind(this));
```

providers/Mora/mod.ts
```diff
@@ -0,0 +1,251 @@
...
+				this.addMessage('Failed to determine GTIN', 'warning');
+			}
+		}
+
+		const release: HarmonyRelease = {
+			title: albumPage.title,
+			artists: [this.makeArtistCreditName(albumPage.artistName)],
+			labels: [label],
+			gtin,
+			releaseDate: parseISODateTime(albumPage.startDate),
```
kellnerd: While this probably works as intended in UTC-N time zones, it returns the previous day in UTC+N time zones (since that function currently only expects the `YYYY-MM-DDTHH:mm:ss.sssZ` format according to its description).
I'm not sure whether we are supposed to interpret the 0:00 as UTC or as Japanese time... If it is UTC, the solution is as simple as the following hack (which we should document as accepted format):
```suggestion
			releaseDate: parseISODateTime(albumPage.startDate + ' +0'),
```
P.S. Generally the date parser utilities are something for which I should write test cases and then refactor them. Your usage of this function (and the SoundCloud PR) show that I can probably handle more formats with the same function if I add a few safeguards on top of the [notoriously surprising JS Date parser](https://jsdate.wtf/).

providers/Mora/mod.ts
```diff
@@ -0,0 +1,251 @@
...
+	convertRawTrack(index: number, rawTrack: Track): HarmonyTrack {
+		return {
+			number: index,
+			title: rawTrack.title,
+			type: rawTrack.mediaFormatNo == MediaFormat.Video ? 'video' : 'audio',
+			artists: [this.makeArtistCreditName(rawTrack.artistName)],
+			length: rawTrack.duration * 1000,
+			recording: {
+				externalIds: [],
+			},
```
kellnerd: This is optional, so unless mora has track IDs and accessible track pages it should be omitted.
```suggestion
```

providers/Mora/mod.ts
```diff
@@ -0,0 +1,251 @@
...
+		};
+	}
+
+	makeArtistCreditName(artist: string): ArtistCreditName {
+		return {
```
kellnerd: Both tracks and packages have an `artistNo` property which we can use to construct artist IDs here.
Untested code:
```suggestion
	makeArtistCreditName(artist: string, artistNo): ArtistCreditName {
		return {
			externalIds: this.provider.makeExternalIds({ type: 'artist', id: artistNo }),
```
(There are also label IDs, but they don't seem to have accessible pages.)

providers/Mora/mod.ts
```diff
@@ -0,0 +1,251 @@
...
+} from '@/harmonizer/types.ts';
+import { type CacheEntry, MetadataProvider, ReleaseLookup } from '@/providers/base.ts';
+import { DurationPrecision, FeatureQuality, FeatureQualityMap } from '@/providers/features.ts';
+import { parseISODateTime, PartialDate } from '@/utils/date.ts';
+import { ProviderError, ResponseError } from '@/utils/errors.ts';
+import { extractMetadataTag } from '@/utils/html.ts';
+import { isValidGTIN } from '../../utils/gtin.ts';
+
+export default class MoraProvider extends MetadataProvider {
+	readonly name = 'Mora';
```
kellnerd:
```suggestion
	readonly name = 'mora';
```
seems to be their preferred spelling?

providers/Mora/mod.ts
```diff
@@ -0,0 +1,251 @@
...
+		hostname: this.supportedUrls.hostname,
+		pathname: '/artist/:id{/}?',
+	});
+
+	override readonly features: FeatureQualityMap = {
+		// The API returns a "full size" image of at most 200x200
+		'cover size': 200,
+		'duration precision': DurationPrecision.SECONDS,
+		'GTIN lookup': FeatureQuality.MISSING,
+		'MBID resolving': FeatureQuality.PRESENT,
```
kellnerd, line 36: Not present yet, but this will work once we construct artist IDs which can be resolved.

providers/Mora/mod.ts
```diff
@@ -0,0 +1,251 @@
...
+					if (apiArgs) {
+						return new Response(JSON.stringify(apiArgs), response);
+					}
+				} catch (_error) {
+					throw new ResponseError(this.name, 'Failed to extract API arguments', webUrl);
+				}
+
+				throw new ResponseError(this.name, 'Failed to extract API arguments', webUrl);
+			},
+		}).then(({ content }) => {
```
kellnerd: I'm not yet sure about the best way to structure the flow with the two dependent requests.
Ideally we would have this second part included in the lookup class, whose instances also store the API base URL. Then we don't have to inject the URL into the JSON object temporarily.
We also need to implement at least the part which appends `packageMeta.json` inside the lookup's `constructReleaseApiUrl` method, whose result is used for display on the website and in edit notes.

kellnerd (changes requested):
Thank you! Despite the number of comments, this is looking good already.
I will soon test the provider a bit myself to see where we potentially need more test cases.

P.S. Regarding the CI complaints about formatting, we can fix that by ignoring the testdata folder in the `fmt.exclude` setting in `deno.json`.

providers/Mora/mod.ts
Serial-ATA: Yeah, I was also searching around for label pages. I don't think they exist.

providers/Mora/mod.ts
Serial-ATA: Is [`3770c02` (#168)](https://github.com/kellnerd/harmony/pull/168/commits/3770c021f821e581421e23ea0b2eb020dc6ce952) what you meant?

providers/Mora/mod.ts
Serial-ATA: Yeah, I've heard JS dates are weird, and since `parseISODateTime()` just delegates to the `Date` parser I figured it would *probably* support the date format on mora (`YYYY/MM/DD HH:mm:ss`).

No idea on the timezone, so I'll just use that hack for UTC.

providers/Mora/mod.ts
kellnerd: Yes, that should be it. You probably have to update the tests' expectations as well.
(Can't see it, CI already fails before it gets to the tests. You can reproduce CI behavior by running `deno task ok` and your new tests.)

providers/Mora/mod.ts
kellnerd: UTC is probably a safe bet, since the time component always seems to be 0:00 and the date component matches the displayed date on the website. So Japanese time (+9) would be a day earlier if we were to interpret that JS Date as UTC.

providers/Mora/mod.ts
Serial-ATA: Updated the tests in [`df55435` (#168)](https://github.com/kellnerd/harmony/pull/168/commits/df55435e79fbe53b5b0e327be069e4ab8aa04b73)

providers/Mora/json_types.ts
```diff
@@ -0,0 +1,52 @@
...
+	master: string;
+	distPartNo: string;
+	startDate: string;
+	mediaFormatNo: MediaFormat;
+	mediaType: MediaType;
+	trackList: Track[];
+}
+
+export interface Track {
+	arranger?: string;
```
kellnerd:
```suggestion
	arranger: string | null;
```
Nitpick: Properties seem to be `null` rather than missing, the same probably also holds for composer and lyrics (but I haven't seen them empty yet, I only encountered `"Not Applicable"`). Did you see any cases where they were missing or did you only use/assume the `?` marker as a safety measure? (In most cases it wouldn't matter whether something is missing, `undefined` or `null`.)

providers/Mora/mod.test.ts
```diff
@@ -0,0 +1,59 @@
...
+import { assertSnapshot } from '@std/testing/snapshot';
+
+import MoraProvider from './mod.ts';
+import { assertEquals } from 'std/assert/assert_equals.ts';
+
+describe('Mora provider', () => {
+	const bc = new MoraProvider(makeProviderOptions());
+	const lookupStub = stubProviderLookups(bc);
+
+	describeProvider(bc, {
```
kellnerd:
```suggestion
	const mora = new MoraProvider(makeProviderOptions());
	const lookupStub = stubProviderLookups(mora);

	describeProvider(mora, {
```

providers/Mora/mod.test.ts
```diff
@@ -0,0 +1,59 @@
...
+		urls: [{
+			description: 'package page',
+			url: new URL('https://mora.jp/package/43000006/00602488058599/'),
+			id: { type: 'package', id: '43000006/00602488058599' },
+			isCanonical: true,
+		}, {
+			description: 'album page with tracking parameter',
```
kellnerd: Since I'm already nitpicking on this file, otherwise I wouldn't bother:
```suggestion
			description: 'package page with tracking parameter',
```

providers/Mora/mod.ts
```diff
@@ -0,0 +1,234 @@
...
+import { ProviderError, ResponseError } from '@/utils/errors.ts';
+import { extractMetadataTag } from '@/utils/html.ts';
+import { isValidGTIN } from '@/utils/gtin.ts';
+
+export default class MoraProvider extends MetadataProvider {
+	readonly name = 'mora';
+
+	readonly supportedUrls = new URLPattern({
+		hostname: 'mora.jp',
+		pathname: '/package/:labelCode([0-9]+)/:materialNo{/}?',
```
kellnerd:
```suggestion
		pathname: '/package/:labelCode([0-9]+)/:packageId{/}?',
```
materialNo is the number in the API URL.

providers/Mora/mod.ts
```diff
@@ -0,0 +1,234 @@
...
+		if (releaseResult) {
+			const { labelCode, materialNo } = releaseResult.pathname.groups;
+			if (!labelCode || !materialNo) {
+				return undefined;
+			}
+
+			return {
+				type: 'package',
+				id: [labelCode, materialNo].join('/'),
+			};
```
kellnerd: Following the previous suggestion:
```suggestion
			const { labelCode, packageId } = releaseResult.pathname.groups;
			if (!labelCode || !packageId) {
				return undefined;
			}

			return {
				type: 'package',
				id: [labelCode, packageId].join('/'),
			};
```

providers/Mora/json_types.ts
Serial-ATA: Just an assumption, I haven't seen a case with it missing

PR #172, opened 2025-11-25
feat(ototoy): add provider for ototoy.jp

### LINKED ISSUE #34: OTOTOY
I honestly couldn't find any API for OTOTOY, but since it is a Japanese store, most of the help pages aren't in English, so there might be.

https://ototoy.jp

that said, perhaps it could be scraped for data, especially since it's one of the few stores I know that shows catalog numbers (for example, [here](https://ototoy.jp/_/default/p/637303)).

important note, OTOTOY does keep seperate pages for Lossless and High-Resolution releases, which would be the same MusicBrainz release (all other data being the same, of course)

Serial-ATA:
I used [jsdom](https://www.npmjs.com/package/jsdom) for the scraping, but can switch to something else if you have a preferred library. I'm not sure what the best option is.

Unfortunately, there's no:

* GTIN support
* Available regions (Not exclusively Japan, https://github.com/kellnerd/harmony/issues/34#issuecomment-2834314687)

Also, there *are* technically URLs for tracks. For example the track for https://ototoy.jp/_/default/p/3016055 is available at https://ototoy.jp/opus/index.php/19511684, but it's just a redirect back to the album page, so I didn't figure it'd be worth adding.

closes #34


kellnerd:
I'm sorry for the delay, I had really hoped to review the provider this week, but it will probably take me until next week.

harmonizer/types.ts
```diff
@@ -103,6 +103,8 @@ export type Artwork = {
 	types?: ArtworkType[];
 	comment?: string;
 	provider?: ProviderName;
+	/** Whether to set `referrerpolicy="no-referrer"` for the artwork URLs */
+	noReferrer?: boolean;
```
kellnerd: Interesting that OTOTOY denies access to images based on the referer header...

I think I would prefer to set `<meta name='referrer' content='same-origin' />` in `server/routes/_app.tsx` rather than setting this per image and per external link.
It shouldn't have any downside to never send a referrer header to external (provider) entity and image links?

providers/mod.ts
```diff
@@ -24,6 +25,7 @@ providers.addMultiple(
 	DeezerProvider,
 	iTunesProvider,
 	SpotifyProvider,
+	OtotoyProvider,
```
kellnerd: I would put this before or after mora. This would keep the HTML extracting providers at the end of the list.

server/icons/BrandOtotoy.tsx
kellnerd, line 1: The icon is too large compared to the other ones. We should shrink it a bit (using a version before the coordinates have been rounded).

providers/Ototoy/mod.ts
```diff
@@ -0,0 +1,460 @@
...
+
+	readonly supportedUrls = new URLPattern({
+		hostname: 'ototoy.jp',
+		pathname: '/_/default/p/:id',
+	});
+
+	readonly artistUrlPattern = new URLPattern({
+		hostname: this.supportedUrls.hostname,
+		pathname: '/_/default/a/:id',
+	});
```
kellnerd, line 25: We can combine those patterns (untested code):
```suggestion
	readonly supportedUrls = new URLPattern({
		hostname: 'ototoy.jp',
		pathname: '/_/default/:type(a|p)/:id(\\d+)',
	});
```
In `extractEntityFromUrl` we would call the parent method and map the letter types to "artist"/"package".

providers/Ototoy/mod.ts
```diff
@@ -0,0 +1,460 @@
...
+					currentDisc = parseInt(match[0], 10);
+				}
+
+				continue;
+			}
+
+			const trackNumberCell = trackRow.querySelector('.num');
+			if (!trackNumberCell) continue;
+
+			const trackNumber = trackNumberCell.textContent?.trim() ?? '';
```
kellnerd: In a couple of places:
```suggestion
			const trackNumber = trackNumberCell.textContent.trim();
```
This doesn't seem to be necessary according to TypeScript?

providers/Ototoy/mod.ts
```diff
@@ -0,0 +1,460 @@
...
+
+			const durationCell = trackRow.querySelectorAll('td')[3];
+			if (!durationCell) continue;
+
+			const duration = durationCell.textContent?.trim() ?? '';
+			if (!duration) return undefined;
+
+			tracks.push({
+				title: title,
+				discNumber: currentDisc ?? undefined,
```
kellnerd:
```suggestion
				discNumber: currentDisc,
```
Not necessary if you initialize the variable with `undefined`.

providers/Ototoy/mod.ts
```diff
@@ -0,0 +1,460 @@
...
+				continue;
+			}
+
+			const trackNumberCell = trackRow.querySelector('.num');
+			if (!trackNumberCell) continue;
+
+			const trackNumber = trackNumberCell.textContent?.trim() ?? '';
+			if (!trackNumber) return undefined;
+
+			const titleSpan = trackRow.querySelector("td.item span[id^='title-']");
```
kellnerd, line 198: It is also possible to extract track artist links from this cell.
See track 2.3 of https://ototoy.jp/_/default/p/3287123 for an example, but there should be releases which are more suitable as test cases.

providers/Ototoy/mod.ts
```diff
@@ -0,0 +1,460 @@
...
+
+		const artistSpans = Array.from(albumMetadata.querySelectorAll('p.album-artist > span.album-artist'));
+		if (artistSpans.length === 0) return undefined;
+
+		const artists: Artist[] = [];
+		for (const span of artistSpans) {
+			const anchor = span.querySelector('a');
+			if (!anchor) return undefined;
+
+			const id = anchor.getAttribute('href')?.match(this.entityPathPattern)?.[1];
```
kellnerd: I would prefer to keep the scraper part simple and directly return the artist URL (or label URL further down).
The ID extraction should happen as part of the release conversion, there is a convenience helper `makeExternalIdsFromUrl` that can be used.
As a bonus we no longer need the regular expressions which are duplicating the URL patterns.

kellnerd (changes requested):
Thank you for doing the first steps with DOM parsing in Harmony!
I was worried that it might be very slow, but in my few measurements (using `performance.now()`) it turned out that fetching the HTML is an order of magnitude slower than parsing and scraping it.
(Example for the Beatles Anthology test case: 323ms parsing / 375ms parsing+scraping of 4751ms total provider time. Smaller releases were more like 100ms of 1500ms.)

providers/Ototoy/mod.ts
```diff
@@ -0,0 +1,460 @@
...
+		albumMeta.label = {
+			name: labelName,
+			id: labelId,
+			catalogNumber,
+		};
+
+		return albumMeta;
+	}
+
+	extractEmbeddedJson<Data>(webUrl: URL, maxTimestamp?: number): Promise<CacheEntry<Data>> {
```
kellnerd: Hmm... we aren't really extracting JSON and the method only returns one specific data type. Maybe
```suggestion
	fetchAndScrapePackagePage(webUrl: URL, maxTimestamp?: number): Promise<CacheEntry<PackagePage>> {
```

providers/Ototoy/mod.ts
Serial-ATA: Yeah, didn't think so. I think my IDE was just bugging

providers/Ototoy/mod.ts
Serial-ATA: Oh good catch, I had no idea. Those links have no highlighting and don't change the cursor :thinking:

server/icons/BrandOtotoy.tsx
Serial-ATA: Is it? It definitely looks out of place since it's filled, but it's pretty much the same size as the MB icon.

I'll shrink it tomorrow if I get a chance, too tired to open Inkscape right now :smile:

kellnerd: You've got a point, the MB icon is also slightly too big.
The version that I had submitted upstream [got altered](https://github.com/tabler/tabler-icons/commit/34cb341cbde7aecbac2346d370f480aa40919661#diff-94b1281dc4e69e403382cc40fb2f51738e7554c4f8ad385d93a9afa1a8992c14) before it was released :sweat_smile:

providers/Ototoy/__snapshots__/mod.test.ts.snap
```diff
@@ -112,12 +111,11 @@ snapshot[`OTOTOY provider > release lookup > multi-disc release 1`] = `
   ],
   images: [
     {
-      noReferrer: true,
-      thumbUrl: "https://imgs.ototoy.jp/imgs/jacket/3237/00000003.3237840.1763733479.356_320.jpg",
+      thumbUrl: "https://imgs.ototoy.jp/imgs/jacket/3237/00000003.3237840.1764965039.7088_320.jpg",
```
kellnerd, line 115: Hmm, I hope the changed timestamp in the URL just indicates that the image of this (newish) release has been replaced, and not that they periodically change all the image URLs.
In that regard it is a good sign that this is the only `thumbUrl` that has changed.

providers/Ototoy/mod.ts
```diff
@@ -0,0 +1,461 @@
...
+			if (!titleSpan) return undefined;
+
+			const artists: Artist[] = [];
+			const artistsHtml = trackRow.querySelectorAll('td.item a.artist');
+			for (const artist of artistsHtml) {
+				const path = artist.getAttribute('href');
+				if (!path) continue;
+
+				artists.push({
+					url: `${this.origin}${path}`,
```
kellnerd:
```suggestion
					url: new URL(path, this.origin).href,
```
Using the URL constructor is probably safer (here and in a few other places).
Since the `path` is root-relative with a leading slash, using just the domain as base URL is fine. If the path was relative, we would need the scraped package URL as base URL here.

providers/Ototoy/mod.ts
```diff
@@ -0,0 +1,461 @@
...
+	// <table id="tracklist" class="tracklist">
+	//    <tbody>
+	//       <tr class="disc-row">
+	//          <td colspan="5">DISC 1</td>
+	//       </tr>
+	//       <tr class="">
+	//          <td class="num center">
+	//             1
+	//          </td>
+	//          <td class="item">
```
kellnerd, line 161: Maybe you want to add a track artist example here?

providers/Ototoy/mod.test.ts
```diff
@@ -0,0 +1,121 @@
...
+					['880492', '1329419'],
+					['880492', '807897', '1258599'],
+				];
+
+				media.tracklist.forEach((track, index) => {
+					const expectedArtists = trackArtists[index].flatMap((value) =>
+						provider.makeExternalIds({ type: 'artist', id: value })
+					);
+					assertEquals(track.artists!.flatMap((artist) => artist.externalIds![0]), expectedArtists);
+				});
```
kellnerd: Just for future reference: You don't have to assert for specific values here which are already covered by the snapshot. Checking that track 6 has 3 artists with an external ID would probably be sufficient.

providers/Ototoy/mod.ts
```diff
@@ -0,0 +1,461 @@
...
+		releaseElements.forEach((el) => {
+			const text = el.textContent.trim();
+			if (text.startsWith('Original release date:')) {
+				originalReleaseDate = text.replace('Original release date:', '').trim();
+			} else if (text.startsWith('Release date:')) {
+				releaseDate = text.replace('Release date:', '').trim();
+			}
+		});
+
+		releaseDate = releaseDate || originalReleaseDate;
```
kellnerd: I wonder whether we should have such a fallback, there are already enough providers returning the "wrong" original release date. Hopefully Harmony will be better equipped to handle release dates in the future and flag the accuracy of date values.
In this regard it would be better if the cached format contains both date values to leave the decision up to the conversion logic later.
Additionally the dates should be treated as optional rather than causing an extraction error.

Serial-ATA: AFAICT, every release has a date. It's just that they either have both an original and a platform release date *or* just one.

https://ototoy.jp/_/default/p/1822344 for example only has an original release date, which I assume means that the original release date == platform release date. (The release on MB even has the same date: https://musicbrainz.org/release/fa036443-0131-4c4a-a838-d9d033184859)

providers/Ototoy/__snapshots__/mod.test.ts.snap
Serial-ATA: Not sure, but both still work so hopefully it's just a revision.

providers/Ototoy/mod.ts
kellnerd: Ah, that makes sense. So we will keep the fallback, but move it into the conversion method and store both extracted dates in the cached data (to give us room for future improvements).

providers/Ototoy/mod.ts
```diff
@@ -291,20 +293,19 @@ export default class OtotoyProvider extends MetadataProvider {
 
 		releaseElements.forEach((el) => {
 			const text = el.textContent.trim();
+			console.log(text);
```
kellnerd: ;)

kellnerd (approved):
One last thing and we are good to go.

providers/Ototoy/mod.ts
Serial-ATA: Good catch, I actually had that there cause I realized that every release used in the tests (except for The Beatles - Anthology Collection) only has an original release date. I wonder if that's the most common pattern.

Issue #181, opened 2026-01-12
Tidal duplicates album artists


phw:
For some reason Harmony started to get double album artist credits from labels, where it ends up with the album artist being "The Artist & The Artist".

E.g. see https://harmony.pulsewidth.org.uk/release?tidal=478959100&spotify=3JHkIBKfmzHOXPrmcDEHno&gtin=4065629759803&deezer=&itunes=&region=GB&ts=1768197015

But in my tests this happened with all releases from Tidal. If there are multiple artists on the album it duplicates each of them, e.g. picking the random https://harmony.pulsewidth.org.uk/release?tidal=16533576&spotify=0DJcrAJy2n8luCPGvxM5sl&gtin=8717306944908&deezer=&itunes=&region=GB&ts=1768199545 gives the credit "Arctic Moon, Arctic Moon, Truewave & Truewave".

I haven't yet checked what changed in the Tidal API to trigger this.

PR #182, opened 2026-01-12
Fix Tidal album artist duplication

phw:
This is a fix for the Tidal artist duplication reported in #181

The existing code returned all matching entries from included items, but Tidal API sometimes returns duplicates. Also the included items not necessarily reflect the same order as the relationships.

I haven't seen the latter being a problem, but purely from the code it is a potential issue that is also fixed with this change. And the existing test case seems to have at least at some point had this case, where it contained "Bruno Mars" and "Lady Gaga" as album artist as opposed this other way around (compare https://tidal.com/album/381265361).

### DIFF
--- a/providers/Tidal/__snapshots__/mod.test.ts.snap
+++ b/providers/Tidal/__snapshots__/mod.test.ts.snap
@@ -442,17 +442,6 @@ snapshot[`Tidal provider > release lookup > single by two artists (v2 API) 1`] =
 {
   artists: [
     {
-      creditedName: "Bruno Mars",
-      externalIds: [
-        {
-          id: "3658521",
-          provider: "tidal",
-          type: "artist",
-        },
-      ],
-      name: "Bruno Mars",
-    },
-    {
       creditedName: "Lady Gaga",
       externalIds: [
         {
@@ -462,6 +451,17 @@ snapshot[`Tidal provider > release lookup > single by two artists (v2 API) 1`] =
         },
       ],
       name: "Lady Gaga",
+    },
+    {
+      creditedName: "Bruno Mars",
+      externalIds: [
+        {
+          id: "3658521",
+          provider: "tidal",
+          type: "artist",
+        },
+      ],
+      name: "Bruno Mars",
     },
   ],
   copyright: "© 2024 Interscope Records",
--- a/providers/Tidal/v2/lookup.ts
+++ b/providers/Tidal/v2/lookup.ts
@@ -349,10 +349,20 @@ export class TidalV2ReleaseLookup extends ReleaseApiLookup<TidalProvider, Single
 		resourceType: 'artists' | 'artworks' | 'providers',
 	): T[] {
 		const targetRels = rawRelease.data.relationships[relationship];
-		const relatedIds = new Set(targetRels?.data?.map((item) => item.id));
+		if (!targetRels) {
+			return [];
+		}
+		const relatedIds = new Set(targetRels.data?.map((item) => item.id));
 
-		return rawRelease.included
-			.filter((resource) => resource.type === resourceType && relatedIds.has(resource.id)) as T[];
+		const items = new Map<string, T>();
+		rawRelease.included.forEach((resource) => {
+			if (resource.type === resourceType && relatedIds.has(resource.id)) {
+				items.set(resource.id, resource as T);
+			}
+		});
+		// Rely on the target relationship to get the correct number and order of related
+		// items. Consider the case that an item was not actually returned in the includes.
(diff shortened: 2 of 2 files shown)


providers/Tidal/__snapshots__/mod.test.ts.snap
```diff
@@ -442,26 +442,26 @@ snapshot[`Tidal provider > release lookup > single by two artists (v2 API) 1`] =
 {
   artists: [
     {
-      creditedName: "Bruno Mars",
+      creditedName: "Lady Gaga",
       externalIds: [
         {
-          id: "3658521",
+          id: "3534754",
```
kellnerd, line 448: Good catch, I didn't even notice the wrong order when I added that test case. But the new order makes sense, it follows the cover art and the Tidal website (as of today at least).

P.S. Have you already tried to update the cached API testdata for this test case? I wonder whether the artist entries are now duplicated as well. I probably wouldn't commit the bloated testdata.
But maybe the Armin van Buuren release with both missing and duplicated resources would make for a good test case?
I guess we could report both API issues to Tidal, maybe they will fix them.

providers/Tidal/v2/lookup.ts
```diff
@@ -349,10 +349,20 @@ export class TidalV2ReleaseLookup extends ReleaseApiLookup<TidalProvider, Single
...
-			.filter((resource) => resource.type === resourceType && relatedIds.has(resource.id)) as T[];
+		const items = new Map<string, T>();
+		rawRelease.included.forEach((resource) => {
+			if (resource.type === resourceType && relatedIds.has(resource.id)) {
+				items.set(resource.id, resource as T);
+			}
+		});
+		// Rely on the target relationship to get the correct number and order of related
+		// items. Consider the case that an item was not actually returned in the includes.
+		return targetRels.data.map((rel) => items.get(rel.id) || { id: rel.id } as T);
```
kellnerd, line 365: Do we currently need the fallback value at all? As far as I can see it doesn't matter whether we return the incomplete object or `undefined`, all callers of `getRelatedItems` aren't prepared to handle either case.
That's the big downside of the type assertion here, you don't notice such issues until they happen.

So far I've only seen missing resources for track artists, which has a completely separate implementation in `getTrackArtists` which doesn't call `getRelatedItems`.
In my local files I've even left a note on this function which mentions the possibility of the current issue, but I hadn't addressed it so far:
> TODO: Track AC conversion is inconsistent with release AC conversion (which might be in the wrong order since it filters the existing resources instead of mapping the IDs to resources).
> Store artistMap as this.artistMap?

In the long term we probably want to properly type the `getRelatedItems` helper or abandon it if there is no clean way to do it. I like the approach with the `artistMap` that we do for tracks, maybe a generic `this.resourceMap` and a suitable helper function is the way to go?

providers/Tidal/__snapshots__/mod.test.ts.snap
phw: No, I haven't tried updating the cached testdata. I don't actually know how to handle this.

But I think I was lucky that the test data exposed this issue, because it allowed testing the situation.

providers/Tidal/v2/lookup.ts
phw: Exactly because of the concerns that also release artists might be missing. I haven't seen such a case, but I think it could happen.

But yes, we could refactor this to a similar system then as used for tracks. Actually the way tracks handle this was added later, and it was done separated as the existing `getRelatedItems` didn't work for this use case.

phw (commented):
I have updated the commit message. But I won't have the time do deeply refactor the artist handling. Hence I would suggest to merge the fix and do the refactoring separately.

> It might break with https://tidal.com/album/16533576, the missing track artist is probably also a release artist which the currently deployed version skips?

No, it doesn't, the release artists are all present. I don't know of a case for which it isn't. But we should expect that this can happen.

kellnerd (approved):

providers/Tidal/__snapshots__/mod.test.ts.snap
kellnerd: I've updated the historic testdata locally, the new response would indeed duplicate the two artists (and add a few new but irrelevant fields as well).
Since this specific test case is also used to guarantee support for an older variant of the v2 API, I didn't commit the changes.

> I haven't tried updating the cached testdata. I don't actually know how to handle this.

https://github.com/kellnerd/harmony/blob/main/CONTRIBUTING.md#testing should get you started the next time.

providers/Tidal/v2/lookup.ts
kellnerd: I've added #185 to track this.

Issue #107, opened 2025-05-13
Add an api for external apps


RustyNova016:
It would be nice to have an API to get the information provided by harmony programmatically. Could be useful to have an userscript that warn missing release links for the release, or recordings, etc...

I'd definitly love a way to have the missing recordings links so that I can directly put an edit link in Alistral's MB Clippy, without making the user go to Harmony first

FoxxMD:
I'm the developer of [multi-scrobbler](https://github.com/FoxxMD/multi-scrobbler) and have been planning on implementing something adjacent to harmony for awhile now, but the [feature has been brought up recently by a user.](https://github.com/FoxxMD/multi-scrobbler/issues/334)

It would be amazing to be able to use harmony as a backend for this feature. A few things that would be essential for me:

* Having JSON endpoints
* Querying by more fields, or a combination of fields, rather than only URL/barcode/MBID
  * Many of the services multi-scrobbler receives data from only present artist, title, and album
* Optional authentication for providers, to allow more usage than anonymous access

PR #190, opened 2026-02-07
Add api

ligeia:
First PR, hopefully I didn't screw up too badly.

This adds an API to Harmony. I've included an API.md file available in project root for review. The PR consists of a slightly modified server/main.ts and a new server/api.ts. I formatted everything with deno fmt and deno task ok shows no issues.

### DIFF
--- a/API.md
+++ b/API.md
@@ -1,0 +1,344 @@
+# Harmony API
+
+A JSON API for querying music metadata from Harmony's aggregated provider data.
+
+## Setup
+
+### Environment Variables
+
+```bash
+HARMONY_API_PORT=5221       # Required: Port to run the API on
+HARMONY_API_HOST=127.0.0.1  # Optional: Host to bind to (default: 127.0.0.1)
+```
+
+### Running
+
+The API starts automatically when `HARMONY_API_PORT` is set:
+
+```bash
+# Via Fresh server (recommended)
+deno task server
+
+# Standalone
+deno task api
+```
+
+---
+
+## Endpoints
+
+All release endpoints accept:
+
+| Parameter | Description |
+|-----------|-------------|
+| `url` | Provider URL (Spotify, Deezer, iTunes, Bandcamp, Beatport, Tidal, etc.) |
+| `gtin` or `barcode` | Barcode/GTIN number |
+| `merge=true` | Optional: Query all providers and merge results (may fail if GTINs conflict) |
+
+---
+
+### `GET /api/health`
+
+Health check endpoint.
+
+```bash
+curl http://localhost:5221/api/health
+```
+
+```json
+{
+  "ok": true,
+  "version": "mvp-20",
+  "timestamp": "2026-01-17T05:21:12.453Z"
+}
+```
+
+---
+
+### `GET /api/v1/barcode`
+
+Quick lookup for barcode and basic release metadata.
+
+```bash
+curl "http://localhost:5221/api/v1/barcode?url=https://open.spotify.com/album/4LH4d3cOWNNsVw41Gqt2kv"
+curl "http://localhost:5221/api/v1/barcode?gtin=886445613261"
+```
+
+```json
+{
+  "mbid": null,
+  "barcode": "886445613261",
+  "title": "The Dark Side of the Moon",
+  "artist": "Pink Floyd",
+  "trackCount": 10,
+  "mediaCount": 1,
+  "source": "harmony"
+}
+```
+
+---
+
+### `GET /api/v1/match`
+
+Candidate matching format — useful for matching workflows.
+
+```bash
+curl "http://localhost:5221/api/v1/match?url=https://open.spotify.com/album/..."
+curl "http://localhost:5221/api/v1/match?gtin=886445613261"
+```
+
+```json
+{
+  "input_url": "https://open.spotify.com/album/...",
+  "input_gtin": null,
+  "candidates": [{
+    "mbid": null,
+    "barcode": "886445613261",
+    "title": "The Dark Side of the Moon",
+    "artist": "Pink Floyd",
+    "trackCount": 10,
+    "mediaCount": 1,
+    "score": 1.0,
+    "evidence": ["harmony-lookup"]
+  }],
+  "best": {
+    "mbid": null,
+    "barcode": "886445613261",
+    "title": "The Dark Side of the Moon",
+    "artist": "Pink Floyd",
+    "trackCount": 10,
+    "mediaCount": 1,
+    "score": 1.0,
+    "evidence": ["harmony-lookup"]
+  },
+  "providers": [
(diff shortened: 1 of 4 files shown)


server/api.ts
```diff
@@ -0,0 +1,434 @@
...
+	// Primary location: release group MBID (populated by MBID resolver)
+	if (release.releaseGroup?.mbid) {
+		return release.releaseGroup.mbid;
+	}
+
+	// Fallback: check if MusicBrainz provider returned data
+	const mbProvider = release.info.providers.find((p) => p.internalName === 'musicbrainz');
+	if (mbProvider?.id) {
+		return mbProvider.id;
+	}
```
kellnerd, line 38: This is wrong, the first is a release group MBID, the second is a release MBID.

server/api.ts
```diff
@@ -0,0 +1,434 @@
...
+ * Extract track count information from a HarmonyRelease.
+ */
+function extractTrackInfo(release: HarmonyRelease): { mediaCount: number; trackCounts: number[]; totalTracks: number } {
+	const trackCounts = release.media.map((m) => m.tracklist.length);
+	const totalTracks = trackCounts.reduce((sum, c) => sum + c, 0);
+	return {
+		mediaCount: release.media.length,
+		trackCounts,
+		totalTracks,
+	};
```
kellnerd, line 66: I feel like this is redundant data and should be computed by the client?

server/api.ts
```diff
@@ -0,0 +1,434 @@
...
+
+				const candidate = {
+					mbid: extractMbid(merged),
+					barcode: extractBarcode(merged),
+					title: merged.title,
+					artist: merged.artists.map((a) => a.creditedName || a.name).join(', '),
+					trackCount: trackInfo.totalTracks,
+					mediaCount: trackInfo.mediaCount,
+					score: 1.0,
+					evidence: ['harmony-lookup'],
```
kellnerd, line 278: What's the purpose of these properties (and the "match" endpoint in general)?

API.md
```diff
@@ -0,0 +1,344 @@
...
+Health check endpoint.
+
+```bash
+curl http://localhost:5221/api/health
+```
+
+```json
+{
+	"ok": true,
+	"version": "mvp-20",
```
kellnerd, line 51: What's this? Shouldn't this rather return the Harmony release version?

server/api.ts
```diff
@@ -0,0 +1,434 @@
...
+		},
+	},
+};
+
+function formatDuration(ms: number): string {
+	const totalSeconds = Math.floor(ms / 1000);
+	const minutes = Math.floor(totalSeconds / 60);
+	const seconds = totalSeconds % 60;
+	return `${minutes}:${seconds.toString().padStart(2, '0')}`;
+}
```
kellnerd, line 406: I don't think that the API should format durations, but if it does, there is already an existing utility for that.

API.md
```diff
@@ -0,0 +1,344 @@
...
+### Running
+
+The API starts automatically when `HARMONY_API_PORT` is set:
+
+```bash
+# Via Fresh server (recommended)
+deno task server
+
+# Standalone
+deno task api
```
kellnerd, line 23: This doesn't exist?

API.md
```diff
@@ -0,0 +1,344 @@
...
+
+- Spotify
+- Deezer
+- iTunes / Apple Music
+- Bandcamp (URL lookup only, no GTIN)
+- Beatport
+- Tidal
+- MusicBrainz
+- mora (URL lookup only, no GTIN)
+- OTOTOY (URL lookup only, no GTIN)
```
kellnerd, line 344: This isn't API specific and shouldn't be mentioned here. The web app already generates this information for the About page.

kellnerd (changes requested):
First off, I really appreciate that you have provided documentation with examples. That makes it a lot easier to understand what you are trying to achieve.

I also like that you have kept the API standalone and independent from the web app, this will make testing and the initial development iterations a lot easier.
Personally I wouldn't integrate the API into the web app until it is ready and the endpoints are reasonably stable.
And if we do that, the API definitely shouldn't start a second web server (on a different port) from the web app's main module, it should be mounted into the existing server of the Fresh app. (Once Harmony's been migrated to Fresh 2.0, composition of middlewares will be a lot easier, similar to Express.js etc.)

I haven't really thought much about how an API for Harmony should look like, so I am open to suggestions by people who would use the API.
Nevertheless I've been confused by two of your decisions:
1. There are many endpoints which effectively do the same thing (a release lookup), but return different parts of the resulting data. I would have thought one endpoint `/api/v1/release/lookup` that returns all data (or accepts parameters) should be sufficient?
Especially `/api/v1/match` looks oddly-specific with the `candidates` and `best` property which always return the same single release.
2. I don't really see the need for `serializeRelease`... The existing `HarmonyRelease` type is already fully JSON-compatible and your serialization deviates only in minor aspects.Most of the differences feel like your personal preferences to me. But if you have well-reasoned improvement suggestions, I would rather change `HarmonyRelease` and have one consistent format.

We should probably go back to #107 to collect ideas for the initial API endpoints where everyone who is interested has a chance to participate?

---

P.S. I also wanted to mention that deploying an API is currently not scalable, but that isn't specific to your implementation. There are mainly two concerns, one is inherent to the concept of Harmony, the other is related to the current architecture:
1. The web app alone is already hitting the limits of the underlying APIs quite frequently nowadays. Hosting an API on the same server would only make this much worse.
2. All the outgoing API requests are currently cached indefinitely in order to guarantee that the permalinks in edit notes continue to work and provide evidence using the original/historic data that was used to import the release.
The `SnapStorage` cache therefore accumulates a lot of (mostly JSON) data (about 57GB in 18 months). While this is acceptable for the Release Lookup page which has a high probability of resulting in a MB import, it is wasteful for Release Actions already and almost pointless for most API requests that I would expect.

So before I feel confident to deploy an API into production, at least the second concern has to be resolved to avoid cluttering the `SnapStorage` with tons of data. Therefore Harmony needs a second, more traditional caching layer which discards data after a certain time (probably 24 hours). Only Release Lookup (or requests which explicitly need a permalink) should write to the permanent `SnapStorage` cache.
If the first concern turns out to be an issue, we can always turn off the API to improve the situation while it is hard to impossible to get rid of the useless API snapshots.

ligeia:
FYI I am testing a newer, simpler version of this that I started from scratch. In the new version, there is a single endpoint (api/release) that receives all the same parameters that the web interface lookups would. It uses only Harmony's existing functions, is integrated into the existing Fresh server, and simply takes what might appear in the web interface and instead provides it as JSON from a GET to the endpoint. This approach is simpler and, I think, more consistent with what users are already used to.

We may need to discuss the caching question but I am hopeful that this is closer to something that would be acceptable.

Issue #101, opened 2025-04-17
Don't seed the release date if it's before the source service existed


Aerozol:
The functionality to filter out release dates if it's before the source service existed.
Harmony internally already knows the start dates of each service. 

Some [notes from kellnerd](https://community.metabrainz.org/t/harmony-music-metadata-aggregator-and-musicbrainz-importer/698641/223?u=aerozol):
- Harmony should not censor source data, the doubtful dates should still be displayed but not seeded. Otherwise I fear that people will either think that Harmony failed to get a date or just copy the doubtful date from somewhere else.
- The current date objects could be extended with a new optional property which explains if/why a date is probably incorrect and will not be used. Maybe we even need an additional “quality” flag because dates from some providers can’t be relied on in general (like iTunes always using the original release date, which is only correct if the digital release happens to be the first release).
- Placeholder dates like January 1st should probably be auto-corrected while still being transparent about this behavior. Not sure if/how this should be combined with the new “explanation” property, maybe this should only happen at the merge stage when the placeholder date is the only remaining plausible date candidate.

jesus2099:
Yes it would be great not importing wrong dates by discarding dates when they are before source site existence: https://community.metabrainz.org/t/stop-importing-wrong-release-dates-from-streaming-sites/814960?u=jesus2099

PR #197, opened 2026-03-21
Provide quality indication with release date

oliversalzburg:
Wraps the provided release date with additional information on the assumed quality of the date. This is primarily used to indicate when a release predates the launch date of the platform.

This change only expands the data structure, without making further use of the newly available information. This was intentionally left open to be resolved in a future change.

Contributes to: #101


oliversalzburg:
Forgot to `deno fmt`. Should be good now.

kellnerd (approved):
Thank you, that's a good start! We probably want to have additional quality levels for providers which you can't generally trust (e.g. because they use the original release date).
But we can add that later, same goes for updating the merge and display logic.

PR #205, opened 2026-04-02
fix(web): Adjust referrer policy so MB is aware seeding is coming from Harmony

julian45:
Since at least when commit b9f174f was present in production, the interstitial MusicBrainz shows for confirmation of seeding data from an external source shows "null" when seeding a release from Harmony. Per @mwiencek, this is likely due to the same-origin referrer policy. By moving to the next-strictest option, we should be able to send the right referrer info to MusicBrainz.

Confirmed that this makes MusicBrainz show the correct source in its interstitial by running a test instance of Harmony from the forked branch and verifying behavior when attempting to seed a release.

### DIFF
--- a/server/routes/_app.tsx
+++ b/server/routes/_app.tsx
@@ -14,7 +14,7 @@ export default defineApp((_req, ctx) => {
 				<meta charset='utf-8' />
 				<meta name='viewport' content='width=device-width, initial-scale=1.0' />
 				<title>Harmony</title>
-				<meta name='referrer' content='same-origin' />
+				<meta name='referrer' content='strict-origin-when-cross-origin' />
 				<meta name='description' content='Music Metadata Aggregator and MusicBrainz Importer' />
 				<meta property='og:description' content='Music Metadata Aggregator and MusicBrainz Importer' />
 				<meta property='og:image' content={logoUrl.href} />


kellnerd (changes requested):
We can't change this header globally as this breaks image loading from OTOTOY. That's the reason why this header was originally added at all.

> Interesting that OTOTOY denies access to images based on the referer header...

> I think I would prefer to set `<meta name='referrer' content='same-origin' />` in `server/routes/_app.tsx` rather than setting this per image and per external link.
> It shouldn't have any downside to never send a referrer header to external (provider) entity and image links?
Interesting that OTOTOY denies access to images based on the referer header...

It is definitely no harm (rather a little privacy gain) not to send no referrer to all images and links on the page. I'm not even sure I would consider MB not showing a referrer a bug...
How have you even noticed this at all now that Harmony is skipping the interstitial automatically?

If we really want to set the referrer for MB seeding specifically, I think we can override the global policy just for the posted form element.

julian45:
> We can't change this header globally as this breaks image loading from OTOTOY. That's the reason why this header was originally added at all.

Drat.

> It is definitely no harm (rather a little privacy gain) not to send no referrer to all images and links on the page. I'm not even sure I would consider MB not showing a referrer a bug...

Strictly speaking, I would consider this a bug since it's a deviation from previous behavior that was consistent across seeders I'd used. But, fair point.

> How have you even noticed this at all now that Harmony is skipping the interstitial automatically?

For me, it consistently shows the interstitial for a brief moment (under normal MB load, definitely <1s, usually <~0.5s) before proceeding to the login page. I'm away from my personal computer at the moment, but maybe I could take a screen recording to show this later?


> If we really want to set the referrer for MB seeding specifically, I think we can override the global policy just for the posted form element.

Makes sense. I would appreciate some pointers on this as I'm not particularly familiar with this sort of web dev; would this be handled somewhere in the neighborhood of `server/islands/ReleaseSeeder.tsx`?

kellnerd:
> I would appreciate some pointers on this as I'm not particularly familiar with this sort of web dev; would this be handled somewhere in the neighborhood of `server/islands/ReleaseSeeder.tsx`?

Yes, that would be the right place. Unfortunately it seems like these referrer policy attributes can only be set on link tags, but not on forms. I tried it, but it had no effect.
Unless we revert back to setting the attribute of every image and image link to the strict policy, the global meta header has to stay.

I'm in favor of keeping the current behavior as by design, it is really an MBS issue that it is trying to show a referrer when there is none. This case also occurs when your browser is configured not to send any referrer information, not only when a website like Harmony explicitly configures this.

julian45:
I'm coming back to this later on and beginning to agree with you on what to do in Harmony and what MBS should do. Hence, I've opened [MBS-14336](https://tickets.metabrainz.org/browse/MBS-14336), which will hopefully lead to better handling on the MBS side.

PR #206, opened 2026-04-03
feat(qobuz): Implement Qobuz provider

### LINKED ISSUE #189: Qobuz implementation
I've been working on a Qobuz implementation and am looking for some guidance.

The Qobuz provider implementation currently works by using the app id presented by the web player in the browser. This raises a few questions or problems.

We could make the app id an environmental variable. Users would have to request their own API access.
We could make the app id an environmental variable with a fallback to a hardcoded app id.
We could make the app id an environmental variable, with a fallback to a hardcoded app id, and add a scraper in cases where the old app id is blocked or stops working.

As you might know, API access is granted by emailing qobuz support to ask for it. That's not something I'm going to do and I suspect most users won't either. But is @kellnerd willing to do that? Or do you prefer one of my above suggestions.

That's my biggest roadblock right now.

An aside: there was some discussion elsewhere about whether/when Qobuz responds when searching the catalog by barcode. No, it mostly does not unless the url/slug includes the barcode. I use a combination of HTML scraping and API calls to do barcode searches but the implementation will be similar to bandcamp in that if you start from the Qobuz link, you've got a decent chance to get a barcode and a release you can merge, but if you start from others you will not likely find the release on Qobuz (unless we incorporate much more advanced html scraping than I've figured out).
Also yes, when merging with releases from other platforms, you have to normalize the GTIN either for Qobuz or for the other platforms as Qobuz uses 13 digits and no one else really does.

Lioncat6:
Closes #189


Lioncat6:
~~Tests seem to fail with a server error 500 for whatever reason, but lookups are running fine outside of tests. Perhaps the test command I'm using is incorrect: `deno test --allow-net --allow-write --allow-env ./providers/Qobuz/mod.test.ts -- --download`

Edit: seems the environment variables aren't loading properly during the test~~

Forgot to specify .env path :p

`deno test --allow-net --allow-write --allow-read --allow-env ./providers/Qobuz/mod.test.ts -- --download --update`

Lioncat6:
Just realized eti isn't included in track titles

providers/Qobuz/mod.test.ts
```diff
@@ -0,0 +1,95 @@
...
+			description: 'label page',
+			url: new URL('https://play.qobuz.com/label/97377'),
+			id: { type: 'label', id: '97377' },
+			isCanonical: true,
+		}, {
+			description: 'label page',
+			url: new URL('https://play.qobuz.com/label/97377'),
+			id: { type: 'label', id: '97377' },
+			isCanonical: true,
+		}, {
```
kellnerd, line 59: Exact duplicates, maybe you wanted to test play and open subdomains here, but they don't exist for labels?

providers/Qobuz/mod.ts
```diff
@@ -0,0 +1,232 @@
...
+	QobuzSearchResponse,
+} from '@/providers/Qobuz/api_types.ts';
+
+const qobuzAppId = getFromEnv('HARMONY_QOBUZ_APP_ID') || '';
+export default class QobuzProvider extends MetadataApiProvider {
+	readonly name = 'Qobuz';
+
+	readonly supportedUrls = new URLPattern({
+		hostname: '(play|www|open).qobuz.com',
+		pathname: '/:region?/:type(artist|album|track|interpreter|label)/:slug?/:id',
```
kellnerd: We can't call the combination of country and language code `:region` here, it will cause problems for other providers which only expect a country code as region. The simplest solution would be to call the group `:locale` and be done with it, but this doesn't provide much value.
Alternatively we can extract two named groups `:region` and `:language`, which makes the region usable for other providers, but requires us to define a mapping between regions and default languages (for incoming regions from other providers or the web input where we don't know the language).

providers/Qobuz/mod.ts
```diff
@@ -0,0 +1,232 @@
...
+
+	protected async getRawRelease(): Promise<QobuzAlbum> {
+		const apiUrl = this.constructReleaseApiUrl();
+		if (this.lookup.method === 'gtin') {
+			const { content: searchResponse, timestamp } = await this.provider.query<QobuzSearchResponse>(apiUrl, {
+				snapshotMaxTimestamp: this.options.snapshotMaxTimestamp,
+			});
+			this.updateCacheTime(timestamp);
+			const matchingAlbum = searchResponse.albums?.items.find((album) =>
+				this.removeLeadingZeros(album.upc) == this.removeLeadingZeros(this.lookup.value)
```
kellnerd: There is a barcode comparison helper `isEqualGTIN` in `utils/gtin.ts` already, please use that one and delete the then unused method here.

providers/Qobuz/mod.ts
```diff
@@ -0,0 +1,232 @@
...
+		};
+
+		return {
+			title: rawRelease.title,
+			artists: this.getAlbumCredits(rawRelease),
+			media: this.getAlbumMedium(rawRelease),
+			releaseDate: this.convertReleaseDate(parseHyphenatedDate(rawRelease.release_date_stream)),
+			copyright: rawRelease.copyright || undefined,
+			status: 'Official',
+			types: [capitalizeReleaseType(rawRelease.release_type || (rawRelease.tracks_count > 1 ? 'album' : 'single'))],
```
kellnerd: I wouldn't introduce track count based heuristics here. None of the other providers does that and if we want to have this, it should be implemented as a general helper function (probably in the post-processing stage).

providers/Qobuz/mod.ts
```diff
@@ -0,0 +1,232 @@
...
+			medium.tracklist.push(this.convertRawTrack(track));
+		});
+		result.push(medium);
+		return result;
+	}
+
+	private convertRawTrack(rawTrack: QobuzPartialTrack): HarmonyTrack {
+		return {
+			title: `${rawTrack.title}${rawTrack.version ? ` (${rawTrack.version})` : ''}`,
+			artists: rawTrack.album ? this.getAlbumCredits(rawTrack.album) : [],
```
kellnerd: If Qobuz doesn't have track artists, we should keep them empty and let Harmony/MB infer the release artists automatically.

providers/Qobuz/mod.ts
```diff
@@ -0,0 +1,232 @@
...
+			if (track.media_number !== medium.number) {
+				result.push(medium);
+				medium = {
+					number: track.media_number,
+					format: 'Digital Media',
+					tracklist: [],
+				};
+			}
+			track.album = album; // pass album info to track for artist credits
+			medium.tracklist.push(this.convertRawTrack(track));
```
kellnerd: This hack won't be needed per my previous comment. (If we would need the album object, it should be a parameter of the `convertRawTrack` method.)

server/routes/icon-sprite.svg.tsx
kellnerd, line 1: Please keep the imports and definitions sorted alphabetically (also in some other modules and the CSS file where you have added Qobuz). This also helps to avoid some potential merge conflicts which arise when multiple changes append stuff at the same place.

server/icons/QobuzOutline.tsx
kellnerd, line 1: Let's keep the icon as is for now, but I think I can try to adapt the Q-disc to follow the Tabler grid and stroke style.

providers/mod.ts
```diff
@@ -30,6 +31,7 @@ providers.addMultiple(
 	BeatportProvider,
 	MoraProvider,
 	OtotoyProvider,
+	QobuzProvider,
```
kellnerd: We can probably move Qobuz between Bandcamp and Beatport.
Following my undocumented "ordering guidelines" (from open APIs over authenticated APIs to HTML scrapers and genre-specific providers), it should be between Tidal and Bandcamp probably, but that would be two black buttons in a row 🙅

providers/Qobuz/mod.ts
```diff
@@ -0,0 +1,232 @@
...
+		return {
+			title: rawRelease.title,
+			artists: this.getAlbumCredits(rawRelease),
+			media: this.getAlbumMedium(rawRelease),
+			releaseDate: this.convertReleaseDate(parseHyphenatedDate(rawRelease.release_date_stream)),
+			copyright: rawRelease.copyright || undefined,
+			status: 'Official',
+			types: [capitalizeReleaseType(rawRelease.release_type || (rawRelease.tracks_count > 1 ? 'album' : 'single'))],
+			packaging: 'None',
+			images: [{ url: this.getMaxImage(rawRelease.image.small), types: ['front'] }],
```
kellnerd: Please assign the small version of the image to the thumbnail property, which is the one that is rendered in the web interface. You can probably wrap the entire images logic in a method rather than just the max image part.

providers/Qobuz/mod.test.ts
```diff
@@ -0,0 +1,95 @@
+import type { ReleaseOptions } from '@/harmonizer/types.ts';
```
kellnerd, line 2:
```suggestion
// Automatically load .env environment variable file (before anything else).
import '@std/dotenv/load';

import type { ReleaseOptions } from '@/harmonizer/types.ts';
```

That will ensure that the tests always load your .env file.

providers/Qobuz/__snapshots__/mod.test.ts.snap
```diff
@@ -0,0 +1,177 @@
...
+      day: 13,
+      month: 5,
+      year: 2024,
+    },
+    quality: "assumed-valid",
+  },
+  status: "Official",
+  title: "Suspend My Belief",
+  types: [
+    "Epmini",
```
kellnerd: That would be an invalid type for MB, you should have a mapping function for those.

providers/Qobuz/__snapshots__/mod.test.ts.snap
```diff
@@ -0,0 +1,177 @@
...
+    },
+  ],
+  copyright: "2024 Honeycomb Records 2024 Honeycomb Records",
+  externalLinks: [
+    {
+      types: [
+        "paid streaming",
+        "paid download",
+      ],
+      url: "https://www.qobuz.com/fr-fr/album/suspend-my-belief-ivycomb-stephanafro/rjrikcvbggy0b",
```
kellnerd: I'm not very familiar with Qobuz URL formats and which one is accepted on MB.
I only noticed that you are seeding the localized www URL here while the provider tests consider the non-localized open URLs without slug as canonical. Is there a reason for that discrepancy?

kellnerd (changes requested):
Thank you very much, this is looking good already!
I've got a couple of comments nevertheless. These are only based on reading the code, I will come back once I've tested the provider with a couple of releases (but that will have to wait until I'm done with my Discogs provider testing).

P.S. The formatter doesn't like your API interface definitions, please fix with `deno fmt`. You can locally check that CI (except tests) passes by running `deno task ok`.

providers/Qobuz/__snapshots__/mod.test.ts.snap
Lioncat6: All qobuz links are currently accepted on MB since there's no cleanup for them. The old www.qobuz links are the only ones displayed in the sidebar currently. I personally prefer the open.qobuz links for a couple reasons, so I'll likely switch to using those.

See https://tickets.metabrainz.org/browse/MBS-13611

providers/Qobuz/mod.ts
Lioncat6: Qoubz doesn't seem to have track level artist credits outside of the track.performers string, which seems to be the same for all tracks on an album.

providers/Qobuz/mod.ts
Lioncat6: Using `locale` group for now...

providers/Qobuz/mod.ts
```diff
@@ -0,0 +1,246 @@
...
+		release: 'album',
+		recording: 'track',
+		label: 'label',
+	};
+
+	readonly releaseLookup = QobuzReleaseLookup;
+
+	override readonly launchDate: PartialDate = {
+		year: 2007,
+		month: 8,
```
kellnerd: Where did you get this date from? [Wikipedia](https://en.wikipedia.org/wiki/Qobuz) says 18 September 2007.
(I only checked because your Naver VIBE PR uses the same date.)

providers/Qobuz/mod.ts
```diff
@@ -0,0 +1,246 @@
...
+	readonly supportedUrls = new URLPattern({
+		hostname: '(play|www|open).qobuz.com',
+		pathname: '/:locale?/:type(artist|album|track|interpreter|label)/:slug?/:id',
+	});
+
+	override readonly features: FeatureQualityMap = {
+		'cover size': 3000,
+		'duration precision': DurationPrecision.SECONDS,
+		'GTIN lookup': FeatureQuality.GOOD,
+		'MBID resolving': FeatureQuality.GOOD,
```
kellnerd: Hmm, given that there are dozens of formats and locales, this should rather be expensive (like iTunes):
```suggestion
		'MBID resolving': FeatureQuality.EXPENSIVE,
```

providers/Qobuz/mod.ts
```diff
@@ -0,0 +1,246 @@
...
+			types: this.mapAlbumType(rawRelease.release_type),
+			packaging: 'None',
+			images: this.getAlbumImage(rawRelease.image.small),
+			labels: [{
+				name: rawRelease.label.name,
+				externalIds: this.provider.makeExternalIds({ type: 'label', id: String(rawRelease.label.id) }),
+			}],
+			externalLinks: [{
+				url: rawRelease.url,
+				types: this.provider.getLinkTypesForEntity(),
```
kellnerd: Is there a reason why you are not using the `streamable` and `downloadable` properties here?

P.S. I thought this was part of my first review already, but apparently I forgot to hit the "Add review comment" button and GH remembered it until now 🤔

kellnerd (commented):
> `deno test --allow-net --allow-write --allow-read --allow-env ./providers/Qobuz/mod.test.ts -- --download --update`

Quick tips:
1. You can abbreviate the allow flags with `-NWRE`.
2. Usually you want either the `--download` or the `--update` flag (or none of them most of the time).
3. `--download` should be rarely enabled after all the testdata has been cached once because responses often cause irrelevant noise for git because of unstable sort order (like your Qobuz search response changing with every commit :/) or irrelevant properties (like popularity) changing their values.
4. You should only `--update` the snapshots if there was an intended change that caused an assertion error, otherwise it will always update and swallow unintended breaking changes.

providers/Qobuz/mod.ts
Lioncat6: I'm bad at months apparently...

providers/Qobuz/mod.ts
Lioncat6: Nvm, I know what you mean, I just haven't implemented logic for it yet :p

providers/Qobuz/__snapshots__/mod.test.ts.snap
Lioncat6: Resolved, but I'm leaving this here for the ticket

kellnerd:
My quick update from your other PR applies here as well: Code-wise this is looking good, but I still have to actually test the provider with a few releases myself.

Lioncat6:
After some testing, label matching seems to work perfectly with qobuz, even when using another provider as the source for the UPC:
image

This is really cool since AFAIK this is the only provider to actually have usable label URLs so far

Lioncat6:
Side note: Was doing some testing and I had an issue where 2 artists were getting assigned the same URL.

kellnerd:
A few additional thoughts after I finally found time to test the implementation a bit:
- open.qobuz.com does not show multiple release artists, but the www subdomain does... Maybe we should standardize on the www URL format (for releases), even if it contains locale and slug? Or maybe have both? Really unsure about that, I will check the MBS ticket again when I have more time.
- A track artist can be extracted from the `track.performer` object ([example](https://open.qobuz.com/album/tuw37t7m9o35a)), multiple artists are only included in the `track.performers` string with the role MainArtist.
- I can't verify play.qobuz.com/label links without an account, so I assume that you have checked them as logged in user.

I already did some changes locally, so please don't push any new commits for now, thanks.

Lioncat6:
@kellnerd From what I can tell, the www subdomain is the best, but qobuz does seem to like making their URLs as confusing as possible... I have made some more recent descoveries on that [ticket](https://tickets.metabrainz.org/browse/MBS-13611) that have changed my mind a bit...

open.qobuz.com links are the only ones that are functional without a slug or region, but are themselves region restricted

www.qobuz links work regardless of the user's region, but the region codes do matter in very specific circumstances:
https://www.qobuz.com/jp-ja/album/-/u483rtr3vmcrb <- works
https://www.qobuz.com/us-en/album/-/u483rtr3vmcrb <- does *not*
slugs on www.qobuz links are simply cosmetic and are _should always be cleaned up to a hypen_ (-) as qobuz does themselves for track titles comprised of all non-latin characters.

kellnerd (approved):
I've pushed a few additional commits, the provider should be more or less ready for the initial release now. Most importantly, the provider is now extracting track artists (first performer with ID, all others name only) and catches 404 errors (which gives a more meaningful error for e.g. JP-exclusive releases).

Two open questions remain:
1. Should we remove HARMONY_QOBUZ_AUTH_TOKEN again or would it be useful to keep this? In that case I want to at least document that it is usually not needed.
2. I've kept `constructUrl` as is for now since we don't have the locale and the slug to construct www URLs in all cases. Should we still attempt to do this at least for release URLs whenever possible? Instead of or in addition to open.qobuz URLs?

For me the API responses seem to always assume fr-fr as the locale (even for my requests from DE), so the release URL input is currently the only reliable source for localized Qobuz (release) URLs. I have code to construct www release URLs from the extracted URL region and language (if available) on a separate branch at b901643679b9c77c5abd4b544d950dbb75ba47a0.
If we want to follow that path, we could even use the region from other providers or Harmony's region input.
A matching language to complete the locale could be inferred from the region in that case. Qobuz has only 26 available regions, so that would be manageable.
We could use the release locale for the artist and label URLs as well, but we don't always have the slug available, so I tend to keep those as open/play.qobuz URLs.

providers/Qobuz/mod.ts
kellnerd: I've implemented the extraction of country and language in 5fd1a462a6a12e01d1dd4514f7174f3e1b3f8608.
Unfortunately it doesn't seem to be possible to define an optional path segment with two named groups in an `URLPattern`.
Hence the simplest solution was to have one combined group `region` again and split it into country and language afterwards :sweat_smile:

kellnerd:
> 1. Should we remove HARMONY_QOBUZ_AUTH_TOKEN again or would it be useful to keep this? In that case I want to at least document that it is usually not needed.

Removed it, we can always add it back if there is a proper use case.

> 2. I've kept `constructUrl` as is for now since we don't have the locale and the slug to construct www URLs in all cases. Should we still attempt to do this at least for release URLs whenever possible? Instead of or in addition to open.qobuz URLs?
> 
> For me the API responses seem to always assume fr-fr as the locale (even for my requests from DE), so the release URL input is currently the only reliable source for localized Qobuz (release) URLs. I have code to construct www release URLs from the extracted URL region and language (if available) on a separate branch at [b901643](https://github.com/kellnerd/harmony/commit/b901643679b9c77c5abd4b544d950dbb75ba47a0).

I've integrated this change as that was also reosarevok's preference for the MBS ticket.

> If we want to follow that path, we could even use the region from other providers or Harmony's region input. A matching language to complete the locale could be inferred from the region in that case. Qobuz has only 26 available regions, so that would be manageable.

Done, the region input is used if Qobuz is available there. This may cause invalid URLs in some cases, but the user should simply use the proper region or a localized URL if that happens.

> We could use the release locale for the artist and label URLs as well, but we don't always have the slug available, so I tend to keep those as open/play.qobuz URLs.

I am doing this for the release label because we always have the slug for it and use it to construct a public www URL instead of the rather useless play URL.
For artists it is better to consistently use open URLs rather than mixing www URLs with proper slug and www URLs with placeholder slugs.

Lioncat6:
Forgot to review, but I do agree with the URL changes.

In my own implementation, I did make `HARMONY_QOBUZ_AUTH_TOKEN` optional instead of removing it but for simplicity's sake removing it is fine I think.

PR #209, opened 2026-04-05
feat(vibe): Add Naver VIBE provider

### LINKED ISSUE #208: Naver VIBE

Example album: https://vibe.naver.com/album/35758991

# API

There are two requests to make:

One for the album: `https://apis.naver.com/vibeWeb/musicapiweb/album/<ALBUM ID>.json`
And one for the tracklist: `https://apis.naver.com/vibeWeb/musicapiweb/album/<ALBUM ID>/tracks.json`

So for the example album:

* https://apis.naver.com/vibeWeb/musicapiweb/album/35758991.json
* https://apis.naver.com/vibeWeb/musicapiweb/album/35758991/tracks.json

## `album`

*Super* basic info here. There's not even a release type.

### Cover art

Like #207, just remove the query parameter from `response.result.album.imageUrl` to get the original image. I haven't tested much, but the originals are pretty consistently 3000x3000.

## `album/tracks`

Again, basic info. Nothing notable.

# Prior art

@Lioncat6 already implemented Naver VIBE support in SAMBL: https://github.com/Lioncat6/SAMBL-React/blob/main/lib/providers/naver.ts

Lioncat6:
Got bored and decided to implement VIBE as well...

Might want to take a look at here in particular: https://github.com/kellnerd/harmony/compare/main...Lioncat6:harmony:vibe?expand=1#diff-994ea9806a6ec65de9b9322a6997658f145c4b9a0fc8e88570a249f0a6f34970R145-R258

Closes https://github.com/kellnerd/harmony/issues/208


Lioncat6:
Some albums with a lot of featuring artists for testing:
`https://vibe.naver.com/album/10304443`
`https://vibe.naver.com/album/638616`
`https://vibe.naver.com/album/3034414`

providers/Vibe/mod.ts
```diff
@@ -0,0 +1,278 @@
...
+		recording: 'track',
+		label: 'label',
+	};
+
+	readonly releaseLookup = VibeReleaseLookup;
+
+	override readonly launchDate: PartialDate = {
+		year: 2007,
+		month: 8,
+	};
```
kellnerd, line 61: Should be [June 11 2018](https://en.wikipedia.org/wiki/Naver_VIBE).
Please be very careful when copy-pasting other provider implementations, it is very easy to introduce mistakes and unused code which makes review/maintenance much harder.
There is a reason I have an empty provider template, although I could probably refine it a bit now that there are more provider implementations which often have the same de-facto standard methods.

providers/Vibe/mod.ts
```diff
@@ -0,0 +1,278 @@
...
+		'GTIN lookup': FeatureQuality.MISSING,
+		'MBID resolving': FeatureQuality.GOOD,
+		'release label': FeatureQuality.PRESENT,
+	};
+
+	readonly entityTypeMap = {
+		artist: ['artist', 'interpreter'],
+		release: 'album',
+		recording: 'track',
+		label: 'label',
```
kellnerd: Copy-pasta?
```suggestion
		artist: 'artist',
		release: 'album',
		recording: 'track',
```

providers/Vibe/mod.ts
```diff
@@ -0,0 +1,278 @@
...
+	readonly supportedUrls = new URLPattern({
+		hostname: 'vibe.naver.com',
+		pathname: '/:type(artist|album|track)/:id',
+	});
+
+	override readonly features: FeatureQualityMap = {
+		'cover size': 3000,
+		'duration precision': DurationPrecision.SECONDS,
+		'GTIN lookup': FeatureQuality.MISSING,
+		'MBID resolving': FeatureQuality.GOOD,
```
kellnerd: Given that I've never seen a VIBE link in MB and only Deezer, Spotify and MB are currently considered as GOOD (= frequently linked and easy to lookup):
```suggestion
		'MBID resolving': FeatureQuality.PRESENT,
```

providers/Vibe/mod.ts
```diff
@@ -0,0 +1,278 @@
...
+	};
+
+	readonly apiBaseUrl = 'https://apis.naver.com/vibeWeb/musicapiweb/';
+
+	constructUrl(entity: EntityId): URL {
+		return new URL([entity.type, entity.id].join('/'), 'https://vibe.naver.com/');
+	}
+
+	override getLinkTypesForEntity(): LinkType[] {
+		return ['paid streaming', 'paid download'];
```
kellnerd, line 70: Is this correct, do they have downloads?

providers/Vibe/mod.ts
```diff
@@ -0,0 +1,278 @@
...
+export default class VibeProvider extends MetadataApiProvider {
+	readonly name = 'Naver VIBE';
+
+	override get internalName(): string {
+		return 'vibe'
+	}
+
+	readonly supportedUrls = new URLPattern({
+		hostname: 'vibe.naver.com',
+		pathname: '/:type(artist|album|track)/:id',
```
kellnerd: If all their IDs are numeric, we should make this stricter:
```suggestion
		pathname: '/:type(artist|album|track)/:id(\\d+)',
```

providers/Vibe/mod.ts
```diff
@@ -0,0 +1,278 @@
...
+		const apiUrl = this.constructReleaseApiUrl();
+		if (this.lookup.method === 'gtin') {
+			throw new ProviderError(this.provider.name, 'GTIN lookups are not supported');
+		} else {
+			const { content: naverResult, timestamp } = await this.provider.query<NaverResponse<NaverAlbumResult>>(apiUrl, {
+				snapshotMaxTimestamp: this.options.snapshotMaxTimestamp,
+			});
+			this.updateCacheTime(timestamp);
+			const release = naverResult.response.result.album;
+			return release;
```
kellnerd: KISS
```suggestion
			const { content, timestamp } = await this.provider.query<NaverResponse<NaverAlbumResult>>(apiUrl, {
				snapshotMaxTimestamp: this.options.snapshotMaxTimestamp,
			});
			this.updateCacheTime(timestamp);
			return content.response.result.album;
```

providers/Vibe/mod.ts
```diff
@@ -0,0 +1,278 @@
...
+	private convertRawArtist(rawArtist: NaverPartialArtist): ArtistCreditName {
+		return {
+			name: rawArtist.artistName,
+			creditedName: rawArtist.artistName,
+			externalIds: this.provider.makeExternalIds({ type: 'artist', id: String(rawArtist.artistId) }),
+		};
+	}
+
+	private async getAlbumMedium(album: NaverAlbum): Promise<HarmonyMedium[]> {
+		const result: HarmonyMedium[] = [];
```
kellnerd, line 168: Note for myself: I should really add a helper for this bog-standard flat tracklist to multiple media conversion... almost every provider does this.

providers/Vibe/mod.ts
```diff
@@ -0,0 +1,278 @@
...
+					ms += Number(segments[1]) * 1000;
+			} else if (segments.length == 1){
+					ms += Number(segments[1]) * 1000;
+			}
+			return ms;
+	}
+
+}
+
+class QobuzResponseError extends ResponseError {
```
kellnerd: Copy-pasta

providers/Vibe/mod.ts
```diff
@@ -0,0 +1,278 @@
...
+	private async getRawTrackCredits(id: number): Promise<NaverResponse<NaverTrackCreditsResult>> {
+		const apiUrl = new URL(`track/${id}/credits.json`, this.provider.apiBaseUrl);
+		const { content: naverResult, timestamp } = await this.provider.query<NaverResponse<NaverTrackCreditsResult>>(apiUrl, {
+			snapshotMaxTimestamp: this.options.snapshotMaxTimestamp,
+		});
+		this.updateCacheTime(timestamp);
+		return naverResult;
+	}
+
+	private getTrackDuration(duration: string): number {
```
kellnerd: Can most likely be replaced by `parseDuration` from utils/time.ts

providers/Vibe/mod.ts
```diff
@@ -0,0 +1,278 @@
...
+		return result;
+	}
+
+	private async getRawTracklist(id: number): Promise<NaverResponse<NaverArtistTracksResult>> {
+		const apiUrl = new URL(`album/${id}/tracks.json`, this.provider.apiBaseUrl);
+		const { content: naverResult, timestamp } = await this.provider.query<NaverResponse<NaverArtistTracksResult>>(apiUrl, {
+			snapshotMaxTimestamp: this.options.snapshotMaxTimestamp,
+		});
+		this.updateCacheTime(timestamp);
+		return naverResult;
```
kellnerd: I would already unwrap the actual track data here:
```suggestion
		const { content, timestamp } = await this.provider.query<NaverResponse<NaverArtistTracksResult>>(apiUrl, {
			snapshotMaxTimestamp: this.options.snapshotMaxTimestamp,
		});
		this.updateCacheTime(timestamp);
		return content.response.result.tracks;
```

providers/Vibe/mod.ts
```diff
@@ -0,0 +1,278 @@
...
+		};
+	}
+
+	private async getRawTrackCredits(id: number): Promise<NaverResponse<NaverTrackCreditsResult>> {
+		const apiUrl = new URL(`track/${id}/credits.json`, this.provider.apiBaseUrl);
+		const { content: naverResult, timestamp } = await this.provider.query<NaverResponse<NaverTrackCreditsResult>>(apiUrl, {
+			snapshotMaxTimestamp: this.options.snapshotMaxTimestamp,
+		});
+		this.updateCacheTime(timestamp);
+		return naverResult;
```
kellnerd: I would already unwrap the actual track credits data here, see above.

providers/Vibe/mod.ts
```diff
@@ -0,0 +1,278 @@
...
+			return ({
+				name: artist.artistName,
+				id: artist.artistId,
+				likeCount: 0,
+				imageUrl: artist.imageUrl || null
+			})
+		})
+	}
+
+	private async getTrackArtists(rawTrack: NaverTrack): Promise<ArtistCreditName[]> {
```
kellnerd, line 227: It's hard to follow what is happening here... I would like to see some test cases for this function to see what it is supposed to do and if it can be written in a more obvious way. For example, I also have the feeling that we don't have to distinguish between participants and artists in the end and can have one function which converts artists with optional join phrases but I don't feel confident to try without having test coverage.

kellnerd (changes requested):
Wow, that was fast...

Originally I wanted to continue working on my Discogs provider today, then I decided to give you some quick initial feedback first, but in the end I found a few more things 😅

providers/Vibe/mod.ts
Lioncat6: Yup, though I need to add validation for which tracks do, since all tracks marked as 'overseas' don't

providers/Vibe/mod.ts
Lioncat6: Where all other providers (minus Apple and tidal apparently) would provide all participating main track credits, Naver does not and instead moves featuring artists exclusively to the /credits endpoint for tracks


See: https://harmony.pulsewidth.org.uk/release?url=&gtin=0692788693392&region=&musicbrainz=&deezer=&itunes=&spotify=&tidal= for the album on other services

The album tracks fetch here: [album/34923420/tracks.json](https://apis.naver.com/vibeWeb/musicapiweb/album/34923420/tracks.json) omits the featuring artists, so they have to be fetched with [track/96143271/credits.json](https://apis.naver.com/vibeWeb/musicapiweb/track/96143271/credits.json) ([track/96143271.json ](https://apis.naver.com/vibeWeb/musicapiweb/track/96143271.json) returns the same thing as the album tracks fetch). But /credits also returns all artist roles, including non-participating roles which can be seen [here](https://apis.naver.com/vibeWeb/musicapiweb/track/41598799/credits.json)

Lioncat6: I should probably make a helper function to generate the join phrases for the artists, since vibe does distinguish between featuring artists and main artists, unlike most other providers

Lioncat6: What the logic is currently doing:


1. Check if the track title has an indicator of featuring artists to attempt to save on extra calls; Can be removed if unreliable `if (!trackFeatVariations.some((fv) => rawTrack.trackTitle.toLocaleLowerCase().includes(fv))) {`
2. Fetch track credits `const trackCredits = await this.getRawTrackCredits(rawTrack.trackId);`
3. Filter out featuring artists `const featuringArtists = roles.filter((role) => role.roleName === '피쳐링').flatMap((role) => role.participantList);`
4. Store the ids of the featuring artists `const featuringIds = featuringArtists.map((artist) => artist.id);`
5. Combine and de duplicate all artists `typescript const uniqueArtists = Array.from(new Map(allArtists.map((artist) => [artist.id, artist])).values());`
6. Use a forEach loop to check the positions of the nearby artists to generate the joinPhrases & map to convertRawParticipant
7. Return credits

I want to make a helper function that replaces the job of step 6

providers/Vibe/mod.ts
Lioncat6: Is there a way to specify link types for recordings?

Every naver track can be downloaded, but not every album can.

Lioncat6: Resolved but keeping this open

kellnerd:
Quick update (since the weekend has already passed and I found no time for your PRs): Code-wise this is looking good, but I still have to actually test the provider with a few releases myself.

Maxr1998:
Tested this the other day to add a new release, worked really well! Is there any other tests you'd like me to do?

providers/Vibe/mod.ts
Maxr1998: This is actually a bit of a tricky one, since, as the article says, Naver Vibe merged with their second product, Naver Music, in 2020. Naver Music existed much longer than Naver Vibe, specifically since 2002, if [this article](https://namu.wiki/w/%EB%84%A4%EC%9D%B4%EB%B2%84%20%EB%AE%A4%EC%A7%81) is to be believed. What's interesting is that the IDs from Naver Music and Naver Vibe are identical, going way back and allowing MusicBrainz to migrate the old links to the Vibe format some time ago: https://tickets.metabrainz.org/browse/MBS-13922

So I'd argue this new provider should have the earlier launch date, even if it's technically not 100% correct for Vibe.

Maxr1998: cc @kellnerd since this is resolved and to make sure you see it.

Lioncat6:
@Maxr1998 Thanks for looking at it! I still need to make that [helper function](https://github.com/kellnerd/harmony/pull/209#discussion_r3037379716), and I need to double check to make sure that all tracks on VIBE with featuring artists have one of [these](https://github.com/kellnerd/harmony/pull/209/changes/BASE..54520f1a7af6d2a3a64e032adf6784f25e34bdb3#diff-994ea9806a6ec65de9b9322a6997658f145c4b9a0fc8e88570a249f0a6f34970R203) strings in the track title.

providers/Vibe/mod.ts
Lioncat6: I'll un-resolve it for now :)

kellnerd: Ok, if there are releases on Vibe which could be as old as from 2002, then I agree with using that as the launch date.

P.S. I get notified about all comments in this repo whether you ping me or not :grinning:
But un-resolving the thread was a good idea, otherwise it is impossible to see this without following a link from an email notification.

kellnerd:
There is no official checklist for new providers yet, but it would probably be good to have one.
Other than the obvious stuff (does it crash? any errors in the displayed/seeded data?) and the automatic provider test cases, I usually check that the lookups are reasonably fast, permalinks are working and how a provider behaves for different kinds of releases (many tracks, multiple artists, API properties which look like they could have unexpected values).
I also have an eye on the size of cached data and if there may be better ways to get the same information or if there are any requests which we can omit.

providers/Vibe/mod.ts
Maxr1998: [The website](https://web.archive.org/web/20020325214340/http://music2.naver.com/) appears to have worked a lot differently back then, but for our purposes, I think using 2002 as a launch date is adequate. The (perhaps slightly later?) album URL format still redirects to Vibe: http://music.naver.com/album/index.nhn?albumId=2

Right, sorry about that. I'll keep it in mind for the future. And yes, thanks for un-resolving the thread!

Maxr1998:
Found an album on Vibe with unavailable tracks that are meant to be CD-exclusive: https://vibe.naver.com/album/37484204 (see https://kpop.fandom.com/wiki/Heart_Byte_:_Legacy for details). This is relatively common with Korean streaming providers, these tracks should be ignored when importing the release. The Bugs! integration already does this filtering by each track's `service_yn` attribute, something similar should be possible for Vibe.

Maxr1998:
As expected, it's an easy fix: [0001-fix-vibe-only-include-tracks-that-are-streamable-or-.patch](https://github.com/user-attachments/files/28613219/0001-fix-vibe-only-include-tracks-that-are-streamable-or-.patch)

Maxr1998:
@Lioncat6 do you want me to take over from here? Would love to get this landed as well, now that Bugs! and Melon are done.

PR #213, opened 2026-05-01
feat(Bugs): Implement Bugs! provider

### LINKED ISSUE #202: Bugs!

Example album: https://music.bugs.co.kr/album/4136672

# API

I don't think there's a public API, this was just determined from the requests made by the mobile app.

**Note that an auth token is required**. You don't have to login, you can just take the one that app uses. I can share it by email if needed.

Album information is split into sections, which need to be explicitly requested and any of them can be omitted.

For the example album, the app posts the following data to https://mapi.bugs.co.kr/music/5/multi/invoke/map:

```json
[
  {
    "args": {
      "album_id": 4136672,
      "result_type": "DETAIL"
    },
    "id": "album"
  },
  {
    "args": {
      "album_id": 4136672
    },
    "id": "album_image"
  },
  {
    "args": {
      "album_id": 4136672,
      "result_type": "LIST"
    },
    "id": "album_track"
  },
  {
    "args": {
      "album_id": 4136672
    },
    "id": "album_artist_role"
  },
  {
    "args": {
      "album_id": 4136672,
      "page": 1,
      "result_type": "LIST",
      "size": 1
    },
    "id": "get_albumreview_by_albumid"
  },
  {
    "args": {
      "album_id": 4136672,
      "page": 1,
      "result_type": "LIST",
      "size": 4
    },
    "id": "album_series"
  },
  {
    "args": {
      "album_id": 4136672,
      "page": 1,
      "result_type": "LIST",
      "size": 4
    },
    "id": "album_relation"
  },
  {
    "args": {
      "album_id": 4136672,
      "page": 1,
      "result_type": "LIST",
      "size": 3
    },
    "id": "album_mv"
  }
]
```

Returning:

```json
{
  "list": [
    {
      "album": {
        "result": {
          // ...
        }
      }
    },
    {
      "album_image": {
        "list": [
          // ...
        ],
      }
    },
    {
      "album_track": {
        "list": [
          // ...
        ],
      }
    },
    {
      "album_artist_role": {
        // ...
      }
    },
    {
      "get_albumreview_by_albumid": {
        "list": [
          // ...
        ],
      }
    },
    {
      "album_series": {
        "list": [
          // ...
        ],
      }
    },
    {
      "album_relation": {
        "list": [
          // ...
        ],
      }
    },
    {
      "album_mv": {
        "list": [
          // ...
        ],
      }
    }
  ],
}
```

AFAICT, the only significant sections are `album` and `album_track`. `album_image` seems to be redundant since the `album` object contains an image path already.

So the request for the example can be shortened to:

```json
[
  {
    "args": {
      "album_id": 4136672,
      "result_type": "DETAIL"
    },
    "id": "album"
  },
  {
    "args": {
      "album_id": 4136672,
      "result_type": "LIST"
    },
    "id": "album_track"
  }
]
```

## What's available

From the example album (with irrelevant fields removed):

### Album

```json
"album": {
  "result": {
    "album_id": 4136672,
    "title": "LOFI Bespoke Serenity",
    "artists": [
        // See "Artists" section below
    ],
    "styles": [
      {
        "style_id": 323,
        "style_nm": "하우스"
      }
    ],
    "image": {
      "path": "/41366/4136672.jpg"
    },
    "release_ymd": "20251230",
    "rights": {
      "streaming": {
        "service_yn": true
      },
      "download": {
        "service_yn": true
      }
    },
    "genres": [
      {
        "svc_type": 20010,
        "svc_nm": "일렉트로닉"
      }
    ],
    "album_tp_nm": "EP(미니)",
    "agency_nm": "NHN벅스",
    "nation_cd": "KOR",
    "descr": "하우스룰즈가 선사하는 프라이빗한 휴식, [LOFI Bespoke Serenity]\n\n하우스룰즈 유튜브 채널에 꾸준히 공개되어 사랑을 받아온 Lo-Fi 음원들이\n마침내 하나의 앨범으로 완성되다.\n하우스룰즈 특유의 세련되고 트렌디한 터치가 더해진 로우파이 사운드,\n기분 좋은 릴렉스가 필요한 당신의 시간을 채워보세요.\n\n\nCREDITS\n\nProduced by Seoro (House Rulez)\nLyrics by Annie\nMusic Composed & Arranged by Seoro (House Rulez)\nKeyboard, Sax by Seoro\nSound Mixed by Daisuke Imagane\nMastered by Adam Grover\nExcutive produced by Aftermoon Music",
    "labels": [
      {
        "label_id": 993144,
        "label_nm": "애프터문뮤직"
      }
    ]
  }
}
```

`agency_nm` seems to be the distributor.

### Tracks

```json
"album_track": {
  "list": [
    {
      "track_no": 1,
      "track_id": 6396459,
      "track_title": "Trust",
      "artists": [
        // Same format as album.artists
      ],
      "disc_id": 1,
      "rights": {
        // Same format as album.rights
      },
      "len": "05:27"
    },
    // ...
  ]
}
```

Track URL format: `https://music.bugs.co.kr/track/<track_id>`

### Artist

There's actually a lot of information on the artist provided in the album API, not sure if any of it could be of use.

In the example album, the artists list looks like:

```json
"artists": [
  {
    "artist_id": 80014177,
    "artist_nm": "하우스 룰즈(House Rulez)",
    "type": {
      "group_yn": true,
      "sex_cd_nm": "남성"
    },
    "image": {
      "path": "/800141/80014177.jpg"
    },
    "genres": [
      {
        "svc_type": 20010,
        "svc_nm": "일렉트로닉"
      }
    ]
  }
]
```

### Artwork

Cover art is available at `https://image.bugsm.co.kr/album/images/<size>/<path-from-api>`. Note that `size` can be any number (the server scales on demand), or `original`. `original` images don't have a consistent size. The example album has a 5000x5000 image, while https://music.bugs.co.kr/album/4137006 has a 12500x12500 image.

Maxr1998:
Bugs! is a South Korean digital streaming service that also offers digital downloads (FLAC or MP3). This provider uses the mobile app's API, which requires an access token that can be obtained from a separate endpoint using the client credentials of the app. The device ID is usually queried from the device the app is running on, but can be any random value as it doesn't seem to be validated. Thus, it's set to a hardcoded value which can optionally be overridden through an environment variable. The brand color and icon have been taken from the website.

To implement this change, I used the help of an AI tool (namely Claude Code using the Sonnet 4.6 model), which created the Bugs! API types from the HAR log captured from my device with HTTP Toolkit, and wrote the initial provider implementation and tests scaffold. Regardless, all code changes have been personally reviewed by me, and I take full responsibility for them. I also selected and wrote the test cases myself.

Inspired by and closes #202.


providers/Bugs/api_types.ts
Maxr1998, line 1: These API types include a lot of unused fields, I'm not sure whether I should keep them for completeness and future reference, or remove them for less noise. Please tell me what you prefer.

providers/Bugs/mod.ts
```diff
@@ -0,0 +1,242 @@
...
+		if (bugsDeviceId) apiUrl.searchParams.set('device_id', bugsDeviceId);
+		const accessToken = await this.cachedAccessToken(this.requestAccessToken);
+		return this.fetchJSON<Data>(apiUrl, {
+			policy: { maxTimestamp: options.snapshotMaxTimestamp },
+			requestInit: {
+				method: 'POST',
+				headers: {
+					'Content-Type': 'application/json; charset=UTF-8',
+					'Authorization': `Bearer ${accessToken}`,
+					'User-Agent': 'Mobile|Bugs|5.07.00|Android|16|SM-F936B|samsung|market|105070000',
```
Maxr1998: We might want to customize this user agent too. In my tests, it wasn't validated, so we could also use Harmony's user agent instead of impersonating the mobile app.

providers/Bugs/mod.ts
```diff
@@ -0,0 +1,242 @@
...
+				},
+				body,
+			},
+		});
+	}
+
+	private async requestAccessToken(): Promise<ApiAccessToken> {
+		const url = new URL('https://secure.bugs.co.kr/api/5/appToken');
+		url.searchParams.set('client_id', 'bugsapp_credentials_android');
+		url.searchParams.set('client_secret', 'd33b!z7xeu');
```
Maxr1998: I'm unsure whether this client ID should be in the source code or not. It's technically public since it's shipped in the APK.

providers/Bugs/mod.ts
```diff
@@ -0,0 +1,242 @@
...
+	override readonly launchDate: PartialDate = {
+		year: 2000,
+		month: 2,
+	};
+
+	readonly apiBaseUrl = 'https://mapi.bugs.co.kr/music/5/multi/invoke/map';
+
+	override get internalName(): string {
+		return 'bugs';
+	}
```
kellnerd: This is not necessary, the base class already implements this normalization.

providers/Bugs/api_types.ts
kellnerd: Let's keep them since the work has already been done (and 62 lines of types are far from excessive).

providers/Bugs/mod.test.ts
```diff
@@ -0,0 +1,102 @@
...
+			id: undefined,
+		}],
+		invalidIds: ['abc', 'not-a-number', '123abc'],
+		releaseLookup: [{
+			description: 'Regular (정규) album',
+			release: new URL('https://music.bugs.co.kr/album/4119156'),
+			options: releaseOptions,
+			assert: async (release, ctx) => {
+				await assertSnapshot(ctx, release);
+				assert(release.media.length === 1, 'Should have one disc');
```
kellnerd: Please use `assertEquals` for the the numeric comparisons, it will show us a diff if those assertions should ever fail. (I know I've also made this mistake myself in some old test cases, but I think I've updated those by now.)

providers/Bugs/mod.ts
```diff
@@ -0,0 +1,242 @@
...
+const IMAGE_BASE = 'https://image.bugsm.co.kr/album/images';
+
+const bugsDeviceId = getFromEnv('HARMONY_BUGS_DEVICE_ID') ?? 'harmony';
+
+export default class BugsProvider extends MetadataApiProvider {
+	readonly name = 'Bugs!';
+
+	readonly supportedUrls = new URLPattern({
+		hostname: 'music.bugs.co.kr',
+		pathname: String.raw`/:type(album|track|artist)/:id(\d+)`,
```
kellnerd, line 35: Do they also have label pages? I'm just asking since those also have numeric IDs which we could then use to resolve them to MB label entities.

providers/Bugs/mod.ts
```diff
@@ -0,0 +1,242 @@
...
+		return ['paid streaming', 'paid download'];
+	}
+
+	async query<Data>(apiUrl: URL, options: ApiQueryOptions): Promise<CacheEntry<Data>> {
+		const albumId = Number(apiUrl.searchParams.get('album_id'));
+		const body = JSON.stringify([
+			{ id: 'album', args: { album_id: albumId, result_type: 'DETAIL' } },
+			{ id: 'album_track', args: { album_id: albumId, result_type: 'LIST' } },
+		]);
+		if (bugsDeviceId) apiUrl.searchParams.set('device_id', bugsDeviceId);
```
kellnerd: I haven't tried it yet, but is it necessary to have device_id is as part of the URL? The research in #202 didn't mention it at least.
(It also didn't mention the album_id, but we would need that in any case to distinguish cached API responses by URL.)

providers/Bugs/mod.ts
kellnerd: Have you experimented what happens if we omit the additional headers below? If we were to use Harmony as the user agent, I would like to get rid of those as well.

providers/Bugs/mod.ts
kellnerd: For Qobuz it is similar with the web app credentials being publicly exposed. The Qobuz provider reads them from environment variables nevertheless, as I feel more comfortable with that approach. So we should do that here as well.

It may be a bit annoying for people who run a Harmony instance having to set those up themselves, but since it only affects specific providers I think this is an acceptable compromise. I haven't made my mind up yet whether I would accept a documentation page which summarizes how to obtain these credentials though.

providers/Bugs/mod.ts
```diff
@@ -0,0 +1,242 @@
...
+	return {
+		year: parseInt(match[1]),
+		month: parseInt(match[2]),
+		day: parseInt(match[3]),
+	};
+}
+
+function mapReleaseType(albumType?: string): ReleaseGroupType[] | undefined {
+	if (!albumType) return undefined;
+	if (albumType.includes('EP') || albumType.includes('미니')) return ['EP'];
```
kellnerd: Why are you using an includes check here, can the type field contain additional text? From a brief look at the committed testdata, it doesn't seem so.
If it is possible, we should use a simple string lookup mapping which would be more performant as well.

kellnerd (commented):
Thank you, this is looking good already! I only have a few comments until I get to test the new provider myself.

providers/Bugs/mod.ts
Maxr1998: I wasn't entirely sure whether it would handle the exclamation mark correctly, but I should have tested that maybe :sweat_smile:
Works perfectly without, I'll remove this.

providers/Bugs/mod.ts
Maxr1998: You're right, you can actually omit the device ID. In my testing yesterday, I started with an access token obtained from a logged-in account, where omitting the device ID would throw an error (they probably use it for enforcing device limits for paid subscriptions). However, with the anonymous "account" access token, which appears to be globally unique, the device ID isn't enforced to be supplied since a device limit wouldn't make sense anyway. I'll remove the device ID then.

And yes, the album ID isn't supposed to be part of the URL but is included so that the caching layer works. I'll add a comment explaining this.

providers/Bugs/mod.ts
Maxr1998: My main motivation was for hardening here, for example, against white space. But since it is JSON, and the album types generally seem to be stable, I'll replace it with direct comparisons just like I did for the `OST` type.

providers/Bugs/mod.ts
Maxr1998: Unfortunately not.

providers/Bugs/mod.ts
Maxr1998: I did some more testing and could indeed remove nearly everything. Behaves less like the mobile app now, but still returns the same data.

providers/Bugs/mod.ts
Maxr1998: Works for me. I'll move the client secret to a config variable then.

Maxr1998:
Thanks for the thorough review! I think I caught & fixed everything. I left your comments open so that you can verify and resolve them yourself. Once the review is complete and before this is merged, I'd like to squash the second commit so that the client secret is removed from the history again.

Maxr1998:
I just rebased the PR to fix the conflicts, please let me know if you need anything else.

kellnerd (approved):
Thank you for your patience, I've finally found the time to test the provider and it's looking solid as far as I can tell.
You can now squash your commits (and reword the slightly outdated message) if you want (and still remember all the details after the long delay, sorry for that).
Otherwise I can also do that. I've made two minor changes locally (setting the new category "digital", and error handling for missing credentials) which I will push after the merge.

P.S. [One of your example releases](https://music.bugs.co.kr/album/590752) made me realize that I haven't given Hangul a proper weight in utils/script.ts yet, which makes the detected scripts 86% Latin and only 14% Hangul. By giving Hangul a weight of 3, we would get more reasonable 67% and 33%. What do you think? These values would then be below the threshold where Harmony seeds the scripts.

Maxr1998:
> Thank you for your patience, I've finally found the time to test the provider and it's looking solid as far as I can tell. You can now squash your commits (and reword the slightly outdated message) if you want (and still remember all the details after the long delay, sorry for that). Otherwise I can also do that. I've made two minor changes locally (setting the new category "digital", and error handling for missing credentials) which I will push after the merge.

No worries, and thanks for your review. I have squashed the two commits now (force-push diff is empty), and reworded the message to mention the client secret and drop the device ID. Had to do some reading in the review comment history to remember what happened to it but thankfully it's all documented :smile:

> P.S. [One of your example releases](https://music.bugs.co.kr/album/590752) made me realize that I haven't given Hangul a proper weight in utils/script.ts yet, which makes the detected scripts 86% Latin and only 14% Hangul. By giving Hangul a weight of 3, we would get more reasonable 67% and 33%. What do you think? These values would then be below the threshold where Harmony seeds the scripts.

That seems fair. I have once read the argument that releases should somewhat bias the most specific language/script, which would be Korean/Hangul in this case. Increasing the weight does exactly that, so I'm all for it.

PR #229, opened 2026-06-10
Fix RTL artists link rendering 

YoGo9:
I tested it on my instance the artists icons are rendered properly. 

image

https://harmony.mybrainz.dev/release?url=https%3A%2F%2Fopen.spotify.com%2Falbum%2F109QLZlDgbLiC1b0P7nJo7&category=preferred

Instead of the current rendering which causes confusion when clicking artist links. 

image

https://harmony.pulsewidth.org.uk/release?url=https%3A%2F%2Ftidal.com%2Falbum%2F529736650&gtin=&region=GB%2CUS%2CDE%2CJP&musicbrainz=&deezer=&itunes=&spotify=&tidal=&qobuz=&beatport=

### DIFF
--- a/server/components/LinkedEntity.tsx
+++ b/server/components/LinkedEntity.tsx
@@ -1,5 +1,4 @@
 import { ProviderIcon } from '@/server/components/ProviderIcon.tsx';
-
 import { musicbrainzTargetServer } from '@/config.ts';
 import type { ResolvableEntity } from '@/harmonizer/types.ts';
 import { providers } from '@/providers/mod.ts';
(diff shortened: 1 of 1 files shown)


arsinclair:
Hmm, but that seems intentional to me. Clicking on the artist name should open it in MusicBrainz. It is probably necessary to guess the artist name script and insert [RTL/LTR marks](https://en.wikipedia.org/wiki/Implicit_directional_marks) where necessary rather than moving the artist text away from the link.

arsinclair:
Actually I found a better way. Just convert the containers to flexboxes and it will be handled automatically.

Please take a look at #233.

YoGo9:
Thank you! I will test it tomorrow locally.

YoGo9:
Just tested #233 

Works great!

PR #232, opened 2026-06-22
fix(Qobuz): exception on no track performers

selfisekai:
Example release: https://open.qobuz.com/album/0825646121182


kellnerd (approved):
Thank you for finding and fixing this!
Apparently this only happens if the tracks have neither featured artists nor other credits like composers. That probably explains why I didn't find such a case in my initial testing.

Issue #169, opened 2025-11-23
Rework date parsing helpers


kellnerd:
Harmony has a few utility functions to parse dates into the `PartialDate` format.
`parseISODateTime` currently only expects the `YYYY-MM-DDTHH:mm:ss.sssZ` format according to its description, but [we realized](https://github.com/kellnerd/harmony/pull/168#discussion_r2554157604) that it would also work for mora and SoundCloud.

We should write test cases for the date helpers and then try to combine them into fewer, less specific functions. The current `parseISODateTime` can probably handle more formats with the same function if I add a few safeguards on top of the [notoriously surprising JS Date parser](https://jsdate.wtf/).

PR #234, opened 2026-06-25
feat(Melon): Implement Melon provider

### LINKED ISSUE #203: Melon

Example album: https://www.melon.com/album/detail.htm?albumId=12402595

# API

Like #202, I dug this out of the mobile app.

When loading a release, the app makes 3 requests:

* https://m2.melon.com/m6/v1/album/contents/list.json?albumId=12402595
* https://m2.melon.com/m6/v3/album/info.json?albumId=12402595
* https://m2.melon.com/m6/v2/album/song/list.json?albumId=12402595

## album/contents/list.json

This probably isn't useful, it just lists any music videos and articles related to the release.

## album/info.json

AFAICT, `SELLCNPY` is the distributor and `PLANCNPY` is the label.

For the `ALBUMTYPE` field, the values I've seen are:

* `싱글` (Single)
* `EP`
* `OST`
* `리믹스` (Remix)
	* I haven't seen a remix album, so maybe this is always `Single + Remix`?
* `정규` (Regular)
* `옴니버스` (Compilation)

### Cover art

The original image can be derived from the `ALBUMIMG` or `ALBUMIMGLARGE` field. Borrowed from https://github.com/qsniyg/maxurl/blob/master/src/userscript.ts:

```js
const albumImgLarge = "https://cdnimg.melon.co.kr/cm2/album/images/124/02/595/12402595_20251121154956_1000.jpg?4dfce9df1f0761eb689cb1e82a2e01df";

const originalImg = albumImgLarge.replace(/(images\/.*\/[^/_]*)((_[^/.]*)_)?(_?[^/._]*)?(\.[^/.?]*)(?:[?/].*)?$/, "$1$3_org$5");
```

From my (limited) testing, the original covers are usually 2000x2000.

## album/song/list.json

Only notable thing is that it provides a list of genres, but at least in the case of the example release, most of the entries aren't actually genres? Seems to just be a dumping ground of tags. The release-level genre tags don't seem to have that problem.

Maxr1998:
Melon is a South Korean digital streaming service. This provider uses the mobile app's API, which is publicly accessible without authentication. The API returns album metadata and track listings via separate endpoints. The brand color and icon have been taken from the website.

To implement this change, I used the help of an AI tool (namely Claude Code using the Sonnet 4.6 model), which created the Melon API types from HTTP responses caputured with curl using the endpoints described in #203, and wrote the initial provider implementation (using the previous Bugs! provider as a reference). Regardless, all code changes have been personally reviewed by me, and I take full responsibility for them. I also selected and wrote the test cases myself.

Assisted-by: Claude:claude-sonnet-4.6

Closes #203.


Maxr1998:
Still plan to do some fine-tuning on the provider implementation and fix the ToDo I left there, but I think it is mostly ready to review already.

providers/Melon/mod.ts
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	private coverArtwork(largeUrl: string): Artwork {
+		const originalUrl = largeUrl.replace(
+			/(images\/.*\/[^/_]*)((_[^/.]*)_)?(_?[^/._]*)?(\.[^/.?]*)(?:[?/].*)?$/,
+			'$1$3_org$5',
+		);
+		return {
+			url: originalUrl,
+			thumbUrl: largeUrl,
```
kellnerd: I would consider using the 500px image for the thumbnail, most of the other provider thumbnails are even only half of that..

providers/Melon/mod.ts
```diff
@@ -0,0 +1,231 @@
...
+} from './api_types.ts';
+
+const API_BASE = 'https://m2.melon.com/m6';
+
+export default class MelonProvider extends MetadataApiProvider {
+	readonly name = 'Melon';
+
+	readonly supportedUrls = new URLPattern({
+		hostname: 'www.melon.com',
+		pathname: String.raw`/:type(album|song|artist)/detail.htm\?\1Id=:id(\d+)`,
```
kellnerd: Interesting, I wouldn't have thought that this works, the search query is a separate property of the URLPattern usually. I have an unpublished provider which extracts the ID from the query that way, but this surely wouldn't allow us to have a regex back reference.

providers/Melon/mod.ts
```diff
@@ -0,0 +1,231 @@
...
+import type {
+	MelonAlbumInfo,
+	MelonAlbumInfoResponse,
+	MelonArtist,
+	MelonDisc,
+	MelonSong,
+	MelonSongListResponse,
+} from './api_types.ts';
+
+const API_BASE = 'https://m2.melon.com/m6';
```
kellnerd:
```suggestion
const API_BASE = 'https://m2.melon.com/m6/';
```
allows us to use the two argument URL constructor.

providers/Melon/mod.ts
```diff
@@ -0,0 +1,231 @@
...
+	query<Data>(apiUrl: URL, options?: ApiQueryOptions): Promise<CacheEntry<Data>> {
+		return this.fetchJSON<Data>(apiUrl, {
+			policy: { maxTimestamp: options?.snapshotMaxTimestamp },
+		});
+	}
+}
+
+export class MelonReleaseLookup extends ReleaseApiLookup<MelonProvider, MelonRawRelease> {
+	constructReleaseApiUrl(): URL {
+		const url = new URL(`${API_BASE}/v3/album/info.json`);
```
kellnerd:
```suggestion
		const url = new URL('v3/album/info.json', API_BASE);
```

providers/Melon/mod.ts
```diff
@@ -0,0 +1,231 @@
...
+
+	protected async getRawRelease(): Promise<MelonRawRelease> {
+		if (this.lookup.method === 'gtin') {
+			throw new ProviderError(this.provider.name, 'GTIN lookups are not supported');
+		}
+
+		const albumId = this.lookup.value;
+
+		const infoUrl = this.constructReleaseApiUrl();
+		const songListUrl = new URL(`${API_BASE}/v2/album/song/list.json`);
```
kellnerd:
```suggestion
		const songListUrl = new URL('v2/album/song/list.json', API_BASE);
```

providers/Melon/mod.ts
```diff
@@ -0,0 +1,231 @@
...
+			artists: raw.albumInfo.ARTISTLIST.map((a) => this.convertRawArtist(a)),
+			releaseDate: this.convertReleaseDate(parseISSUEDATE(raw.albumInfo.ISSUEDATE)),
+			labels: this.extractLabels(raw.planCnpy),
+			images: [this.coverArtwork(raw.albumInfo.ALBUMIMGLARGE)],
+			types: mapReleaseType(raw.albumType),
+			availableIn: ['KR'],
+			status: 'Official',
+			packaging: 'None',
+			media: this.buildMedia(raw.cdList),
+			// TODO: check if we can check for download availability
```
kellnerd: I see you already had the same question as me 😀

providers/Melon/api_types.ts
```diff
@@ -0,0 +1,55 @@
...
+	SONGNAME: string;
+	ARTISTLIST: MelonArtist[];
+	/** Duration in seconds as a string */
+	PLAYTIME: string;
+	/** Track number as a string */
+	TRACKNO: string;
+	ISSERVICE: boolean;
+	/** Whether the song is marked as a title song, which is usually actively promoted, has an MV, and/or was previously released as a single. */
+	ISTITLESONG: boolean;
+	ISHOLDBACK: boolean;
```
kellnerd, line 62: Have you seen examples where ISSERVICE and ISHOLDBACK indicate unavailable tracks/releases or something else which is of interest?
There is also an ISFREE attribute in the testdata which you haven't used here.

providers/Melon/mod.ts
```diff
@@ -0,0 +1,231 @@
...
+		return {
+			name: artist.ARTISTNAME,
+			creditedName: artist.ARTISTNAME,
+			externalIds: this.provider.makeExternalIds({ type: 'artist', id: artist.ARTISTID }),
+		};
+	}
+
+	private extractLabels(planCnpy?: string): Label[] {
+		if (!planCnpy) return [];
+		return planCnpy.split(', ').map((name) => ({ name: name.trim() }));
```
kellnerd: We already have a utility function `splitLabels` which should be reused. Feel free to enhance its test cases with a Korean script example.

providers/Melon/mod.ts
```diff
@@ -0,0 +1,231 @@
...
+}
+
+interface MelonRawRelease {
+	albumInfo: MelonAlbumInfo;
+	albumType?: string;
+	planCnpy?: string;
+	cdList: MelonDisc[];
+}
+
+function parseISSUEDATE(date: string): PartialDate {
```
kellnerd, line 213: We should revisit this helper as part of #169 one day 😇

providers/Melon/mod.ts
```diff
@@ -0,0 +1,231 @@
...
+}
+
+function mapReleaseType(albumType?: string): ReleaseGroupType[] | undefined {
+	if (!albumType) return undefined;
+	if (albumType === '싱글') return ['Single'];
+	if (albumType === '정규') return ['Album'];
+	if (albumType === 'EP') return ['EP'];
+	if (albumType === '옴니버스') return ['Compilation'];
+	if (albumType === 'OST') return ['Soundtrack'];
+	if (albumType === '리믹스') return ['Single', 'Remix'];
```
kellnerd: Could you confirm the theory that this is only used for remix singles and not remix albums?

server/icons/BrandMelon.tsx
```diff
@@ -0,0 +1,22 @@
...
+			id='brand-melon'
+			width={size}
+			height={size}
+			viewBox='0 0 24 24'
+			fill={color}
+			{...props}
+		>
+			<path
+				fillRule='evenodd'
+				d='M3.89 15.25A8.0 7.95 0 1 0 19.89 15.25A8.0 7.95 0 1 0 3.89 15.25ZM8.34 15.25A3.55 3.84 0 1 0 15.44 15.25A3.55 3.84 0 1 0 8.34 15.25ZM15.3 3.7A2.95 2.95 0 1 0 21.2 3.7A2.95 2.95 0 1 0 15.3 3.7Z'
```
kellnerd: Nice and simple logo.
I do roughly understand SVG path syntax and wonder whether it is intentional that the logo does not consist of perfect circles (`A8.0 7.95` is x and y radius for example) or a conversion/rounding error at some point.
Not worth changing that right now, one day I may make Tabler-style icons for some of the providers and submit them upstream like I did for the MusicBrainz icon.

kellnerd (commented):
Thank you for giving this a shot. Sorry that it took quite a while till I came back to reviewing it, I needed a short break after your last PR and then it got too hot and my motivation was melting like ice in the sun.

Without having tried the provider so far, here is some initial feedback.

providers/Melon/mod.ts
Maxr1998: It appears that `splitLabels` only works for splitting labels separated by slashes, not commas like Melon uses. Should I make the split character customizable in `splitLabels` then?

server/icons/BrandMelon.tsx
Maxr1998: I retraced the icon from their full name logo, which doesn't use perfect circles. However, I'll look if I can find an actual icon SVG in their Android app. That one seems more circle-ish to me when checking again.

providers/Melon/api_types.ts
Maxr1998: I've been looking into this for a bit after submitting the PR, but couldn't find any release with the `ISHOLDBACK` set to false. I also couldn't find any release that wasn't available both for streaming and purchase. I updated the code for now to simply mark every release as streamable and downloadable, since that's also the default used with URL normalization in MusicBrainz right now.

providers/Melon/mod.ts
Maxr1998: See below.

providers/Melon/mod.ts
Maxr1998: I admit that part was clanker output, but I didn't question it since it worked perfectly. The idea is that Melon always uses the same keyword for the path and the query parameter, and they should never mismatch. We don't really have to enforce that, though, so I can rewrite that part to eliminate the backreference if you like.

providers/Melon/mod.ts
kellnerd: Oh, sorry, I had overlooked that. If Melon specifically uses the comma as a separator, then it makes sense to keep this separate.

P.S. We could probably split on slash and comma in the general utility function even, given that this seems to be fine in case of my copyright parser script and release label fields are even simpler than copyright texts (where the comma is also used to structure the sentence).

providers/Melon/mod.ts
kellnerd: No, I actually like the fact that the back-reference makes the pattern stricter.
I would have to have a deeper look into the specs or docs for URLPattern to see if the behavior of "query as part of the path" is expected or depending on the implementation, but I guess it is fine.

providers/Melon/mod.ts
Maxr1998: Sure! Should I make the required changes within this PR, or should that be a separate change later on? I can also do this on a separate branch/PR and then rebase this PR.

providers/Melon/mod.test.ts
```diff
@@ -0,0 +1,94 @@
...
+		}],
+		invalidIds: ['abc', 'not-a-number', '123abc'],
+		releaseLookup: [{
+			description: 'Regular (정규) album with unavailable CD-only tracks',
+			release: new URL('https://www.melon.com/album/detail.htm?albumId=13691104'),
+			options: releaseOptions,
+			assert: async (release, ctx) => {
+				await assertSnapshot(ctx, release);
+				assertEquals(release.media.length, 1, 'Should have one disc');
+				assertEquals(release.media[0].tracklist.length, 11, 'Should have 11 tracks (without the two CD-only tracks)');
```
kellnerd: I was very confused about these CD-only tracks, because when I visit the website today, it doesn't markup any of the 13 tracks specially. Harmony also shows and imports all 13 tracks.
Then I've compared the live API data with your test snapshot and saw that the ISSERVICE attribute of the last two tracks changed from false to true...

I don't know if these ISSERVICE=false tracks are always temporary and whether it would make sense to import them? Or is this an edge case and there are other examples where the tracks stay unavailable?

providers/Melon/mod.ts
kellnerd: This URL pattern did not actually work in my standalone tests. The provider only works because you are overriding `extractEntityFromUrl` and you use only the domain part of the pattern there. As you can see in the table on the About page, the whole thing gets treated as a path in reality (and back-reference seems not be supported at all):
```ts
> supportedUrls.pathname
"/:type(album|song|artist)/detail.htm%3F1Id=:id(\\d+)"
> supportedUrls.exec("https://www.melon.com/album/detail.htm?albumId=13693526")
null
> supportedUrls.exec("https://www.melon.com/album/detail.htm%3F1Id=13693526")
{
  inputs: [ "https://www.melon.com/album/detail.htm%3F1Id=13693526" ],
  protocol: { input: "https", groups: { "0": "https" } },
  username: { input: "", groups: { "0": "" } },
  password: { input: "", groups: { "0": "" } },
  hostname: { input: "www.melon.com", groups: {} },
  port: { input: "", groups: { "0": "" } },
  pathname: {
    input: "/album/detail.htm%3F1Id=13693526",
    groups: { type: "album", id: "13693526" }
  },
  search: { input: "", groups: { "0": "" } },
  hash: { input: "", groups: { "0": "" } }
}
```

So you should change this to use `URLPattern.search` instead, here is an example for a site which always uses `detail=123` as first query parameter and optionally has additional parameters:
https://github.com/kellnerd/harmony/blob/bbf74ef5ad169b60c8d1dea4db8be0ae9898aa17/providers/Hoerspielforscher/mod.ts#L27-L31
https://github.com/kellnerd/harmony/blob/bbf74ef5ad169b60c8d1dea4db8be0ae9898aa17/providers/Hoerspielforscher/mod.ts#L53-L65
You can probably still assert that the type from the query is the same type as in the path.

providers/Melon/mod.ts
kellnerd: You can do it in the same PR as long as you keep the commits separate in the end.

providers/Melon/mod.test.ts
Maxr1998: Oops, that is indeed an edge case and unexpected for me as well. The agency has been promoting this album for quite some time with multiple music videos dropped long after the initial release, so it seems they also decided to make these exclusive tracks available now. That means I'll have to find a different sample for the tests, I'll probably use this one to reduce the confusion: https://www.melon.com/album/detail.htm?albumId=12416346

It also means that I need to create a new release on MusicBrainz :laughing:

providers/Melon/mod.ts
Maxr1998: Sounds good! I'll also squash my fix-ups into the main commit once I'm done.

providers/Melon/mod.ts
Maxr1998: Thanks for checking, and apologies for committing slightly broken code :sweat_smile:
I'll migrate to URLPattern.search then.

server/icons/BrandMelon.tsx
Maxr1998: Didn't find an SVG, but at least a higher-quality bitmap logo, which is indeed using circles (see below). I retraced that now and updated the logo.

icon_launcher

providers/Melon/mod.ts
Maxr1998: The sample remix albums I found were all `옴니버스`, so “Omnibus” = Compilations.

But I haven't been able to find a single marked as `리믹스` either - I only added it because of the ticket mentioning it. @Serial-ATA do you still remember any of the remix singles you found?

providers/Melon/mod.test.ts
Maxr1998: Fixed it. Now uses https://www.melon.com/album/detail.htm?albumId=12416346.

providers/Melon/mod.ts
Maxr1998: Done: https://github.com/kellnerd/harmony/pull/234/commits/e2647a33c22b7fea9775a758557dd02561c6747a

Once you have looked at the new commits, I'll squash & rebase everything. Just tell me once you're ready.

providers/Melon/mod.ts
Serial-ATA: Here's a few:

* https://www.melon.com/album/detail.htm?albumId=11850165
* https://www.melon.com/album/detail.htm?albumId=11001252
* https://www.melon.com/album/detail.htm?albumId=10518206

When I was looking into #203 I wasn't able to find any `리믹스` that _weren't_ singles. But now I'm seeing both singles and albums under that type, so :shrug:

providers/Melon/mod.ts
Maxr1998: Done, fixed this too. Had to set `idPattern` too because the base provider parser only searches the pathname.

providers/Melon/mod.ts
Maxr1998: Awesome, thanks for the examples and context! So should we update the matching then? Or is defaulting to `Single` still safe?

server/static/harmony.css
kellnerd, line 1: Ah, I guess the [new CSS formatter in Deno 2.9](https://deno.com/blog/v2.9#deno-fmt) behaves a bit differently. I've just verified that older versions won't try to undo this and fail CI.

providers/Melon/mod.test.ts
```diff
@@ -0,0 +1,112 @@
...
+		}],
+		invalidIds: ['abc', 'not-a-number', '123abc'],
+		releaseLookup: [{
+			description: 'Compilation (베스트) album with unavailable CD-only tracks',
+			release: new URL('https://www.melon.com/album/detail.htm?albumId=12416346'),
+			options: releaseOptions,
+			assert: async (release, ctx) => {
+				await assertSnapshot(ctx, release);
+				assertEquals(release.media.length, 1, 'Should have one disc');
+				assertEquals(release.media[0].tracklist.length, 23, 'Should have 23 tracks (without the two CD-only tracks)');
```
kellnerd: Minor:
```suggestion
				assertEquals(release.media[0].tracklist.length, 23, 'Should have 23 tracks (without the CD-only track)');
```

providers/Melon/mod.ts
```diff
@@ -0,0 +1,223 @@
...
+} from '@/harmonizer/types.ts';
+import type {
+	MelonAlbumInfo,
+	MelonAlbumInfoResponse,
+	MelonArtist,
+	MelonDisc,
+	MelonSong,
+	MelonSongListResponse,
+} from './api_types.ts';
+import { splitLabels } from '../../utils/label.ts';
```
kellnerd:
```suggestion
import { splitLabels } from '@/utils/label.ts';
```
Some updates ago, Deno stopped using the import map rule rather than those relative parent directory paths... I'm on Deno 2.7.6 for some other reason and I also have to rewrite and move these auto-imports all the time 😒
I should check again if I can update now as this should be [fixed in version 2.7.10](https://github.com/denoland/deno/issues/29464) apparently.

providers/Melon/mod.ts
kellnerd: The example links are only for remix singles, but as far as I understood @Serial-ATA, we should update the matching and not assume that all are singles (as there are also remix albums using that type).

kellnerd (approved):
Looking good, you can now rearrange the commits while you do the last minor changes.
Especially the test data changes should be squashed to remove them from history.

providers/Melon/mod.ts
Maxr1998: Right, I'll change it then.

server/static/harmony.css
Maxr1998: I moved the CSS formatting to a separate commit.

Maxr1998:
Done. I applied your suggestions as well as a few minor fixes from another review pass with Claude (Fable 5, added to the commit message). I also moved the CSS formatting change to a separate commit, and added support for mobile URLs and live albums.

[See here](https://github.com/kellnerd/harmony/compare/e3e2065436c8ebe6abfd3200310796d0c4ea3309..7ff8df8113a180d5737a96cc7f97c33ba8febcda).

PR #236, opened 2026-07-19
fix(Qobuz): restore missing GTIN check digits

### LINKED ISSUE #228: Qobuz returns invalid UPC with missing check digit
https://harmony.pulsewidth.org.uk/release?url=https%3A%2F%2Fwww.qobuz.com%2Fgb-en%2Falbum%2Fbuild-me-up-from-bones-sarah-jarosz%2F0001589171735&gtin&region=GB%2CUS%2CDE%2CJP&qobuz

The above is not an isolated case because for all the time I've been adding Qobuz release links manually, I kept seeing this on older Qobuz releases where the UPC is incorporated into the URL (before they introduced their own native release IDs).

If you try to submit it, MB will immediately report the invalid barcode. You'll have to manually add the missing check digit 7 at the end.
Furthermore, the current limitation prevents such Qobuz links to get harmonized with equivalents on other vendors for the same release in question.

https://harmony.pulsewidth.org.uk/release?url=https%3A%2F%2Fmora.jp%2Fpackage%2F43000006%2F00015891717357%2F&gtin&region=US&ototoy=156359&deezer&spotify&tidal=21989366&qobuz=0001589171735&beatport

floze-the-genius:
## Summary

- normalize legacy Qobuz UPC values that contain 13 payload digits but omit the GTIN-14 check digit
- preserve Qobuz identifiers that already have a valid GTIN checksum
- add focused regression coverage for both missing and valid check digits

Qobuz historically used the truncated UPC as the album ID for older releases. When that 13-digit value has an invalid checksum, this appends the calculated check digit before exposing it as the release GTIN. Valid identifiers are returned unchanged.

Addresses #228

## Validation

Run with the repository CI version, Deno 2.4.5:

- `deno fmt --check` (`179` files)
- `deno lint` (`172` files)
- `deno task check`
- `deno test --deny-net --allow-read --allow-env` (`37` passed, `598` steps)
- `deno task build`
- focused `deno test -RE providers/Qobuz/mod.test.ts` (`28` steps)

### DIFF
--- a/providers/Qobuz/mod.test.ts
+++ b/providers/Qobuz/mod.test.ts
@@ -6,7 +6,7 @@ import { assert } from 'std/assert/assert.ts';
 import { afterAll, describe, it } from '@std/testing/bdd';
 import { assertSnapshot } from '@std/testing/snapshot';
 
-import QobuzProvider from './mod.ts';
+import QobuzProvider, { normalizeQobuzGtin } from './mod.ts';
 import { assertEquals } from 'std/assert/assert_equals.ts';
 
 describe('Qobuz provider', () => {
@@ -127,6 +127,16 @@ describe('Qobuz provider', () => {
 			'https://play.qobuz.com/label/97377',
 			'Region should be ignored for label without slug, URL would be invalid',
 		);
+	});
+
+	describe('GTIN normalization', () => {
+		it('appends a missing check digit to legacy 13-digit UPC values', () => {
+			assertEquals(normalizeQobuzGtin('0001589171735'), '00015891717357');
+		});
+
+		it('preserves UPC values which already have a valid check digit', () => {
+			assertEquals(normalizeQobuzGtin('0198884774947'), '0198884774947');
+		});
 	});
 
 	afterAll(() => {
--- a/providers/Qobuz/mod.test.ts
+++ b/providers/Qobuz/mod.test.ts
@@ -130,11 +130,11 @@ describe('Qobuz provider', () => {
 	});
 
 	describe('GTIN normalization', () => {
-		it('appends a missing check digit to legacy 13-digit UPC values', () => {
+		it('appends a missing check digit to invalid 13-digit barcodes, turning them into valid GTIN-14', () => {
 			assertEquals(normalizeQobuzGtin('0001589171735'), '00015891717357');
 		});
 
-		it('preserves UPC values which already have a valid check digit', () => {
+		it('preserves GTIN-13 values which already have a valid check digit', () => {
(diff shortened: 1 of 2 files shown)


kellnerd:
Thank you for picking this up. I kind of left my feedback in #228 since I would like to confirm with the Qobuz experts what the proper solution should look like.

kellnerd:
@floze-the-genius Do you also want to address the [second part of the issue](https://github.com/kellnerd/harmony/issues/228#issuecomment-5152022533) or do you prefer to get this merged as-is?

floze-the-genius:
Thanks — I’d prefer to keep #236 focused and merge it as-is. The inbound lookup side is a separate behavior change: it needs an agreed fallback order for 14→13-digit formatting versus retrying the invalid truncated value, plus provider fixtures to avoid false matches. Let’s keep #228 open for that follow-up; I changed `Fixes #228` to `Addresses #228`.

providers/Qobuz/mod.ts
```diff
@@ -29,6 +29,14 @@ import { ResponseError as SnapResponseError } from 'snap-storage';
 
 const qobuzAppId = getFromEnv('HARMONY_QOBUZ_APP_ID') || '';
 
+export function normalizeQobuzGtin(upc: string): string {
```
kellnerd, line 33:
```suggestion
/** Appends the missing check digit to invalid 13 digit barcodes from older Qobuz releases. */
export function normalizeQobuzGtin(upc: string): string {
```

providers/Qobuz/mod.test.ts
```diff
@@ -129,6 +129,16 @@ describe('Qobuz provider', () => {
 		);
 	});
 
+	describe('GTIN normalization', () => {
+		it('appends a missing check digit to legacy 13-digit UPC values', () => {
```
kellnerd:
```suggestion
		it('appends a missing check digit to invalid 13-digit barcodes, turning them into valid GTIN-14', () => {
```
Qobuz may call the field UPC, but UPC is usually a 12 digit barcode. EAN-13 or GTIN-13 would be the correct term, but since we are talking about invalid codes here, I will just call it barcode.

providers/Qobuz/mod.test.ts
```diff
@@ -129,6 +129,16 @@ describe('Qobuz provider', () => {
...
+			assertEquals(normalizeQobuzGtin('0001589171735'), '00015891717357');
+		});
+
+		it('preserves UPC values which already have a valid check digit', () => {
```
kellnerd:
```suggestion
		it('preserves GTIN-13 values which already have a valid check digit', () => {
```

kellnerd (approved):

PR #241, opened 2026-08-05
feat(lookup): Replace labels matching the artist name with [no label]

po5:
See [MusicBrainz guide guidelines](https://musicbrainz.org/doc/Style/Unknown_and_untitled/Special_purpose_label#About_auto-releases_or_self-releases).

### DIFF
--- a/harmonizer/release_label.test.ts
+++ b/harmonizer/release_label.test.ts
@@ -20,6 +20,25 @@ describe('cleanupBogusReleaseLabels', () => {
 		});
 	});
 
+	it('replaces self-published label matching the artist name with [no label]', () => {
+		const label: Label = {
+			name: 'Test Artist',
+			catalogNumber: '12345',
+			externalIds: [],
+		};
+		cleanupBogusReleaseLabels([label], [{ name: 'Test Artist' }]);
+		assertEquals(label, {
+			...noLabel,
+			catalogNumber: '12345',
+		});
+	});
+
+	it('replaces a label matching the credited artist alias', () => {
+		const label: Label = { name: 'Test Artist Alias', externalIds: [] };
+		cleanupBogusReleaseLabels([label], [{ name: 'Test Artist', creditedName: 'Test Artist Alias' }]);
+		assertEquals(label, noLabel);
+	});
+
 	const distroKidPlaceholders = [
 		'Distro Kid', // Tidal
 		'123456 Records DK',
--- a/harmonizer/release_label.ts
+++ b/harmonizer/release_label.ts
@@ -1,17 +1,28 @@
-import { Label } from '@/harmonizer/types.ts';
+import { ArtistCreditName, Label } from '@/harmonizer/types.ts';
 import { noLabel } from '@/musicbrainz/special_entities.ts';
 
 /** Placeholder label names that are used by DistroKid. */
 export const DISTRO_KID_PATTERN = /^(Distro ?Kid|\d+ Records DK\d*)$/;
 
+function replaceByNoLabel(label: Label) {
+	label.name = noLabel.name;
+	label.mbid = noLabel.mbid;
+	delete label.externalIds;
+}
+
 /** Tries to clean up common cases of release labels which are not considered imprints by MusicBrainz. */
-export function cleanupBogusReleaseLabels(labels: Label[]) {
+export function cleanupBogusReleaseLabels(labels: Label[], artists?: ArtistCreditName[]) {
+	const artistNames = artists?.map(({ name, creditedName }) => creditedName ?? name);
+
 	for (const label of labels) {
-		if (label.name && DISTRO_KID_PATTERN.test(label.name)) {
+		// Labels are not considered an imprint if the release was self-published.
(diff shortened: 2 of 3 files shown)


chaban-mb:
Just noting sometimes real labels appear in release artists. E.g. 
`https://harmony.pulsewidth.org.uk/release?spotify=1temg9hXAhOnpdEo1CAhwe&qobuz=yhekwetz6hx2b&gtin=199330794014&deezer=&itunes=&tidal=&region=US&ts=1785993872`

~~I'd love to provide you with more test cases but I don't have more available right now.~~
This tag has more: https://musicbrainz.org/tag/label%20as%20artist

BTW, I have also such a feature in my [Harmony enhancements userscript](https://github.com/chaban-mb/userscripts/blob/37f13418476eeac21fd691a6d90e03c0d2a5984b/src/Harmony%20Enhancements.user.js#L1978-L2055).
However it also suffers from the same problem when it's ambiguous whether a label is an artist or vice versa.

<details><summary>Details</summary>
<p>

### High-Level Summary

* **[PR kellnerd/harmony#241](https://github.com/kellnerd/harmony/pull/241)**: Implements basic self-published label cleanup in Harmony by replacing any release label whose name **exactly matches** one of the credited release artist names (or credited aliases).
* **[`setNoLabel`](file:///d:/Documents/git/userscripts/src/Harmony%20Enhancements.user.js#L1978)**: Implements a significantly more flexible matching strategy in the userscript. It accounts for track-level artists, joint/collaboration artist names separated by feature/collaboration tokens (e.g. `Artist A & Artist B`), case-insensitivity, pre-existing MBIDs, and UI label consolidation.

---

### Detailed Comparison Breakdown

| Feature / Behavior | PR `kellnerd/harmony#241` | `setNoLabel` in `Harmony Enhancements` |
| :--- | :--- | :--- |
| **Matching Strategy** | **Exact Equality (`Array.includes`)**<br>Checks `artistNames?.includes(label.name)`. | **Fuzzy / Iterative Substring Removal**<br>Removes artist names iteratively and checks if remaining chars are empty or only separators (`&`, `x`, `feat.`, `,`, etc.). |
| **Collaborative / Multi-Artist Labels** | ❌ **Misses joint self-releases**<br>If release has artists `[Artist A, Artist B]` and label is `"Artist A & Artist B"`, PR 241 fails to match. | ✅ **Handles joint self-releases**<br>Regex strips `Artist A` and `Artist B`, leaving ` & `, which matches `SEPARATORS_REGEX`. |
| **Artist Sources** | Release artists only (`release.artists`). | Release artists + all track artists across media (`release.artists` + `track.artists`). |
| **Case Sensitivity** | Case-sensitive (`String.includes`). | Case-insensitive (`new RegExp(escapedName, 'i')`). |
| **Existing MBID Safeguard** | ❌ **No MBID check**<br>Replaces the label even if `label.mbid` was already assigned. | ✅ **MBID check**<br>Skips evaluation if `originalLabel.mbid` is already present (`if (originalLabel.mbid) return ...`). |
| **Multi-Label Consolidation** | Replaces each matching label individually with `[no label]`. | If *all* labels are self-releases and count > 1, consolidates them into a single `[no label]` entry in data state and UI. |

---

### Key Observations & Edge Cases

1. **Collaboration Labels (e.g., `Artist 1 & Artist 2`)**:
   * PR 241 will fail when digital stores populate the label field with a combined artist string like `"Artist A & Artist B"` if `release.artists` contains individual artist objects for `"Artist A"` and `"Artist B"`.
   * `setNoLabel` handles this using `SEPARATORS_REGEX` (`/^(?:x|&|,|\/|-|\+|feat\.|ft\.|pres\.|presents|vs\.|vs|[\s\u200B\uFEFF])+$/i`).

2. **Case Differences**:
   * PR 241 will miss label entries with subtle casing differences (e.g., label `"ARTIST NAME"` vs artist `"Artist Name"`).

3. **Pre-Existing MBIDs**:
   * In PR 241, if Harmony's lookup or provider metadata assigned a specific label MBID that happens to match the artist's name, PR 241 will overwrite `label.mbid` with `noLabel.mbid`. `setNoLabel` guards against this.

</p>
</details>

arsinclair:
It is possible to discover if there's an actual label on MB pointing to a particular bandcamp page. If yes, in that case we probably shouldn't set it to `[no label]` and go with the old logic.

kellnerd (changes requested):
Thank you for the PR and the test cases @po5, and thank you for the feedback @chaban-mb and @arsinclair.

The feedback which @chaban-mb hides in the details pretty much covers everything I could have brought up and I would like to see those suggestions addressed. Especially the idea to "remove" artist names from the label name and see what remains and the safeguard for labels that could be matched against MB and therefore have an MBID (the Bandcamp example by @arsinclair is covered by this).

Regarding the opposite scenario of labels in the artist credits: Unfortunately we can only catch those if the label has an MBID (and could potentially "fix" the artist credit instead), which is rarely the case with the existing sources.
In all other cases, Harmony at least shows the original values of the label field for all providers since we are only overwriting them with [no label] at the post-merge stage.

PR #244, opened 2026-08-10
fix(web): show navigation on homepage

### LINKED ISSUE #230: Add a link to /settings to the homepage
The homepage doesn't have the "Harmony / Release Lookup / Release Actions / Settings" navbar, so the only way to access the settings page is to go to the URL directly, or get there from a lookup page.
I think it makes sense to let people change the default region for example, before performing a lookup.

floze-the-genius:
## Summary

- show the existing navigation bar on the homepage
- make Settings, Release Lookup, and Release Actions available before the first lookup
- keep the current navigation component, links, and styling unchanged

Fixes #230

## Validation

- `deno task ok`
- `deno test -RE` — 38 tests / 617 steps passed
- `deno task build` — 8 routes / 5 islands built
- local SSR smoke: `/` and `/settings` return 200, and `/` contains the `/settings` link

### DIFF
--- a/server/routes/_app.tsx
+++ b/server/routes/_app.tsx
@@ -6,7 +6,6 @@ import { defineApp } from 'fresh/server.ts';
 export default defineApp((_req, ctx) => {
 	// OpenGraph image URL must be absolute.
 	const logoUrl = new URL('/harmony-logo.svg', ctx.url);
-	const isLandingPage = ctx.url.pathname === '/';
 
 	return (
 		<html lang='en'>
@@ -24,7 +23,7 @@ export default defineApp((_req, ctx) => {
 				<link rel='manifest' href='/site.webmanifest' />
 			</head>
 			<body>
-				{!isLandingPage && <NavigationBar />}
+				<NavigationBar />
 				<ctx.Component />
 				<Footer />
 			</body>


kellnerd (changes requested):
The navigation bar is intentionally hidden on the landing page:
- 3 out of 4 links are useless, their content is already available on the landing page.
- It looks a bit ugly to have a navigation bar with a small logo directly above the banner with the big logo.

Originally it was also intended that the (advanced) Settings are not visible for first time users, but admittedly I've found myself often navigating from the landing page -> About -> Settings, since I usually want to change settings before initiating a lookup.
So I agree that it makes sense to link to Settings directly from home, but as long as there is only one useful navigation bar item, I would rather show the link elsewhere.

I always wanted to have a short explanation ("How does it work?" / "About Harmony") at the end of the landing page, which would be a good fit for a prominent link to the settings.
Maybe we can start with an info message box "Did you know that you can customize Harmony’s behavior?" where customize is a link to the settings.

kellnerd:
Alternatively we could modify the navigation bar conditionally to hide at least the initial "Harmony" link with the logo on the landing page.

Screenshot with both options

What do people think? @afrocatmusic and @aerozol

Aerozol:
Hmm, I know the links are redundant but I do think it's consistent and 'reassuring' to have the same nav bar at the top. We are creatures of habit.

Issue #159, opened 2025-10-30
Wrong track count for Apple music


fabi321:
For some reason, Harmony reports that there is one track less than all other platforms. But I see the correct track count both in web as well as using the Apple Music API.
I am honestly lost as to why that is.
https://harmony.pulsewidth.org.uk/release/actions?release_mbid=7b0baf97-022c-4636-83a7-ce1cfd1189c3

fabi321:
For an API key, see https://github.com/kellnerd/harmony/issues/5#issuecomment-3467696424

I am not sure, if Harmony is doing requests on the server side or client side. If it does them server side, that would fix the issue.

I'm not stupid enough to pay 100$/yr for an api key ;)

kellnerd:
Thanks, that looks useful. API requests are currently performed on the server side, so that is going to help for local development at least.
I am only a bit worried about using a scraped token for a public, server-side app which causes significantly more traffic originating from the same IP address.

Earlier this year I was discussing the topic with @rphsoftware and they said that the official developer API and the inofficial API used by the Apple Music web app are similar enough to support both with one implementation.
So ideally Harmony's implementation would support both and the public server would use official credentials. I also wouldn't pay that price myself, but there have been offers to donate existing credentials to Harmony (or even pay the costs for it!) in the past.

fabi321:
From my testing, I can use both apis (https://amp-api.music.apple.com/v1/ and https://api.music.apple.com/v1/) with this token. I am personally currently using `amp-api`, but more as a precaution, and the library I'm using is designed for `api`.

It also looks like it is intended for "the frontend", so it is supposed to get some traffic.
The thing I'm more worried about is the stability of it. So far it works, but I believe that might change once they redeploy their frontend.

PR #252, opened 2026-08-16
feat(Apple Music): Add catalog API provider for streaming-only releases

dwei30:
## Summary
- Adds a separate **Apple Music** provider that looks up `music.apple.com` albums via the catalog API (`/v1/catalog/{storefront}/albums/{id}`), including streaming-only releases that the iTunes Search API returns with **no tracks** (and often no GTIN from artwork URLs).
- Prefers the **official** Apple Music API (`api.music.apple.com`) when `HARMONY_APPLE_MUSIC_TOKEN` is set. For self-hosted instances, unofficial AMP (`amp-api.music.apple.com`) is available via `HARMONY_APPLE_MUSIC_AMP_TOKEN` or `HARMONY_APPLE_MUSIC_SCRAPE=true`.
- **iTunes is unchanged** (Search API only). Apple Music is registered only when one of those env vars is set, and only then does it claim `music.apple.com` URLs. `itunes.apple.com` stays on iTunes. Without credentials, Harmony behaves as today.

Auth (pick **one**; first match wins):
1. `HARMONY_APPLE_MUSIC_TOKEN` — official developer JWT
2. `HARMONY_APPLE_MUSIC_AMP_TOKEN` — unofficial AMP JWT
3. `HARMONY_APPLE_MUSIC_SCRAPE` — scrape a MusicKit JWT from `music.apple.com` (self-hosted)

Needs testing with an official Apple Music API token as I don't have access to one.
### `.env`
Copy from `.env.example`. You only need one of these; leave the others blank.
```env
# 1. Official Apple Music API (preferred). MusicKit developer JWT.
HARMONY_APPLE_MUSIC_TOKEN=
# 2. Unofficial AMP JWT, used only if HARMONY_APPLE_MUSIC_TOKEN is unset.
HARMONY_APPLE_MUSIC_AMP_TOKEN=
# 3. Scrape a MusicKit JWT from music.apple.com when no token is set (self-hosted only).
# Set true or false.
HARMONY_APPLE_MUSIC_SCRAPE=
```
Local / self-hosted without an Apple Developer key:
```env
HARMONY_APPLE_MUSIC_SCRAPE=true
```

### Sources
- Related: [iTunes API returns less tracks than Apple Music API](https://github.com/kellnerd/harmony/issues/159)
- AMP catalog / MusicKit token approach: [ToadKing/apple-music-barcode-isrc](https://github.com/ToadKing/apple-music-barcode-isrc)


kellnerd:
Hi @dwei30, thank you for working on this often requested feature.
What I've seen so far looks good to me, but for a big change like this I need more time to test which I didn't have so far. I'm on the move right now and hope to get back to this PR within the next two weeks.

PR #255, opened 2026-08-30
refactor(util): Move parseYYYYMMDD from Bugs! provider to date utils

Maxr1998:
Allows the helper to be reused by the upcoming Genie provider.


Maxr1998:
I remember you mentioning this in the review and feel like it would make sense to split the function move from the next provider PR. I decided to only parse full dates in that function. If you want me to support `YYYY` and `YYYYMM` too, please let me know.

kellnerd (approved):

